PR #10, opened 2022-11-26
feat: fixed issues with adding file to target

xtreem88:
https://github.com/customerio/issues/issues/8669
This PR fixes the issue with xcode not recognizing `PushNotification.swift` and requiring an extra step of adding the file after  `prebuild`


xtreem88:
**Pull request title looks good 👍!** 

If this pull request gets merged, **it will cause a new release** of the software. Example: If this project's latest release version is `1.0.0`. If this pull request gets merged in, the next release of this project will be `1.1.0`. This pull request **is not** a breaking change.

All merged pull requests will *eventually* get deployed. But some types of pull requests will trigger a deployment (such as features and bug fixes) while some pull requests will wait to get deployed until a later time. 

**To merge this pull request, add the label `Ready to merge` to this pull request and I'll merge it for you.**


<details >
<summary>This project uses a special format for pull requests titles. Expand this section to learn more (expand by clicking the ᐅ symbol on the left side of this sentence)...</summary>
<br>

This project uses a special format for pull requests titles. Don't worry, it's easy!

This pull request title should be in this format:

```
<type>: short description of change being made
```

**If your pull request [introduces breaking changes](https://web.archive.org/web/20220725195319/https://nordicapis.com/what-are-breaking-changes-and-how-do-you-avoid-them/)** to the code, use this format:

```
<type>!: short description of breaking change
```

where `<type>` is one of the following:

- `feat:` - A feature is being added or modified by this pull request. Use this if you made any changes to any of the features of the project.
- `fix:` - A bug is being fixed by this pull request. Use this if you made any fixes to bugs in the project. 

- `docs:` - This pull request is making documentation changes, only.
- `refactor:` - A change was made that doesn't fix a bug or add a feature.
- `test:` - Adds missing tests or fixes broken tests.
- `style:` - Changes that do not effect the code (whitespace, linting, formatting, semi-colons, etc)
- `perf:` - Changes improve performance of the code.

- `build:` - Changes to the build system (maven, npm, gulp, etc)
- `ci:` - Changes to the CI build system (Travis, GitHub Actions, Circle, etc)

- `chore:` - Other changes to project that don't modify source code or test files.
- `revert:` - Reverts a previous commit that was made.


### Examples:
```
feat: edit profile photo
refactor!: remove deprecated v1 endpoints
build: update npm dependencies
style: run formatter 
```

Need more examples? Want to learn more about this format? [Check out the official docs](https://www.conventionalcommits.org/).

**Note:** If your pull request does multiple things such as adding a feature _and_ makes changes to the CI server _and_ fixes some bugs then you might want to consider splitting this pull request up into multiple smaller pull requests.

</details>

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -18,15 +18,7 @@ const TARGETED_DEVICE_FAMILY = `"1,2"`;
...
-    bundleIdentifier,
-    bundleShortVersion,
-    bundleVersion,
-    iosPath,
-    appName,
-    iosDeploymentTarget,
-  } = options;
+  const { iosPath, appName } = options;
 
   const projPath = `${iosPath}/${appName}.xcodeproj/project.pbxproj`;
```
levibostian, line 23: The `xcodeproj` directory can have whatever name it wants, not necessarily `$appName`. 

Maybe query the file system to find the name of the directory instead "*.xcodeproj"?

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -235,3 +249,33 @@ const updateNseInfoPlist = (payload: {
...
 };
+
+async function addPushNotificationFile(
+  options: CustomerIOPluginOptionsIOS,
+  xcodeProject: any
+) {
+  const { iosPath, appName } = options;
+  const file = 'PushNotification.swift';
+  const getTargetFile = (filename: string) =>
+    `${iosPath}/${appName}/${filename}`;
```
levibostian, line 260: The source code of the iOS app doesn't need to be `${iosPath}/${appName}/`. The directory can be named something completely different. 

Perhaps a more stable way to find the name of the source code directory is to query the file system for a directory that contains `Info.plist` file or some Swift/objective-c source code?

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -235,3 +249,33 @@ const updateNseInfoPlist = (payload: {
...
+
+  if (!FileManagement.exists(getTargetFile(file))) {
+    FileManagement.mkdir(`${iosPath}/${appName}`, {
+      recursive: true,
+    });
```
levibostian, line 265: I believe this code is saying, "If the source code of the iOS app does not exist in the directory path `$iosPath/$appName`, create the directory". 

Do we want to do that? Force customers to store some source code in a random directory they may not want to use in their iOS app?

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -207,6 +91,136 @@ export const withCioNotificationsXcodeProject: ConfigPlugin<
...
+  xcodeProject.addBuildPhase(
+    [],
+    'PBXFrameworksBuildPhase',
+    'Frameworks',
+    nseTarget.uuid
+  );
+
+  // Edit the Deployment info of the target
+  const configurations = xcodeProject.pbxXCBuildConfigurationSection();
+  for (const key in configurations) {
```
levibostian, line 203: I think that this code is iterating through all targets of the Xcode project and modifying the build settings for it. 

I don't think that's what we want to do. We only want to modify build settings for targets that we created.

src/ios/withNotificationsXcodeProject.ts
xtreem88: `$appName` Is predefined in the config for any Expo RN project, these files are "generated" by Expo and usually have similar structures

src/ios/withNotificationsXcodeProject.ts
xtreem88: This is a target we have created

src/ios/withNotificationsXcodeProject.ts
xtreem88: The aim here is to upload the file into a specific (predefined) directory, and it is expected that `${iosPath}/${appName}/` should exist

src/ios/withNotificationsXcodeProject.ts
xtreem88: This path should exist; the idea here is if for any reason, this code runs before the directory is created, then we want to create it and drop the file in there.

src/ios/withNotificationsXcodeProject.ts
levibostian: How so? From reading the code `xcodeProject.pbxXCBuildConfigurationSection()` it looks like it's getting the build configuration for the whole project instead of 1 individual target?

src/ios/withNotificationsXcodeProject.ts
levibostian: Is it a problem if the path does not exist? That is what my original comment is meant to ask.

src/ios/withNotificationsXcodeProject.ts
levibostian: Is it a problem if the path does not exist and the directory is named something other then `$appName`? That is what my original comment is meant to ask.

src/ios/withNotificationsXcodeProject.ts
levibostian: Ok. If the files are generated by Expo which means they have similar structure then, yes, we can assume that this is the naming pattern. 

If Expo didn't generate these files then I would say that might not be the case. 

I'll close my other comments. 

It might be a good idea to add a comment to the code saying that Expo generates these files so we can assume the naming pattern. As an iOS developer, this would answer questions that I would have.

src/ios/withNotificationsXcodeProject.ts
xtreem88: Yes, the code iterates through all the targets but only makes changes to the target we have created here:
```
if (
      typeof configurations[key].buildSettings !== 'undefined' &&
      configurations[key].buildSettings.PRODUCT_NAME ===
        `"${CIO_NOTIFICATION_TARGET_NAME}"`
    ) {
```

levibostian: Ohhhhhhh. I see that now.

levibostian:
@xtreem88 Why is this PR a breaking change? 

Since we are still just in `alpha`, I don't see any issues with introducing breaking changes to the code without any migration steps taken. However, I am still curious how this is a breaking change?

xtreem88:
> @xtreem88 Why is this PR a breaking change?
> 
> Since we are still just in `alpha`, I don't see any issues with introducing breaking changes to the code without any migration steps taken. However, I am still curious how this is a breaking change?

Thanks for pointing this out. This was a mistake... I was testing the integration with `pr_lint` job, and I forgot to change this. I'll do that now.

xtreem88:
@ami-aman @levibostian I am going to clean up this PR soon. I have been testing the changes with different Expo managed projects, and it seems to hold up well. After this, I would need another round of reviews.

src/helpers/constants/ios.ts
```diff
@@ -1,6 +1,6 @@
 export const LOCAL_PATH_TO_CIO_NSE_FILES = `node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
 export const IOS_DEPLOYMENT_TARGET = '13.0';
-export const CIO_SDK_VERSION = '1.2.6';
+export const CIO_SDK_VERSION = '2.0.0-alpha.1';
```
xtreem88, line 3: I am updating the SDK version so that I can test it with recent versions of `customerio-reactnative`

xtreem88: @levibostian Do you think it's a good idea to leave this SDK version and ship it as is?

ami-aman: Segun, is there a way to auto pull in the latest version of iOS SDK / CocoaPods ? Same for react native package version

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -209,6 +100,137 @@ export const withCioNotificationsXcodeProject: ConfigPlugin<
...
+    bundleVersion,
+    bundleShortVersion,
+    infoPlistTargetFile,
+  });
+
+  // Create new PBXGroup for the extension
+  const extGroup = xcodeProject.addPbxGroup(
+    files,
+    CIO_NOTIFICATION_TARGET_NAME,
+    CIO_NOTIFICATION_TARGET_NAME
```
ami-aman, line 153: Same variable twice ? 🤔

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -237,3 +259,33 @@ const updateNseInfoPlist = (payload: {
...
+  options: CustomerIOPluginOptionsIOS,
+  xcodeProject: any
+) {
+  const { iosPath, appName } = options;
+  const file = 'PushService.swift';
+  const getTargetFile = (filename: string) =>
+    `${iosPath}/${appName}/${filename}`;
+
+  if (!FileManagement.exists(getTargetFile(file))) {
+    FileManagement.mkdir(`${iosPath}/${appName}`, {
```
ami-aman, line 273: I see this code being used at multiple places, it would be a good idea to move it to a function ?

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -209,6 +100,137 @@ export const withCioNotificationsXcodeProject: ConfigPlugin<
...
+  // Xcode projects don't contain these if there is only one target
+  // An upstream fix should be made to the code referenced in this link:
+  //   - https://github.com/apache/cordova-node-xcode/blob/8b98cabc5978359db88dc9ff2d4c015cba40f150/lib/pbxProject.js#L860
+  const projObjects = xcodeProject.hash.project.objects;
+  projObjects['PBXTargetDependency'] = projObjects['PBXTargetDependency'] || {};
+  projObjects['PBXContainerItemProxy'] =
+    projObjects['PBXTargetDependency'] || {};
+
+  if (xcodeProject.pbxTargetByName(CIO_NOTIFICATION_TARGET_NAME)) {
+    console.warn(
```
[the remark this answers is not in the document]
xtreem88: This manipulates xcode project asynchronously, so at this point, the new target has not been saved; this block checks if the target already exists and then stops processing it does.

src/ios/withNotificationsXcodeProject.ts
ami-aman: Gotcha. How about adding a comment here to explain the same ? To me it looked like a mistake that the same property name has been passed twice.

src/ios/withNotificationsXcodeProject.ts
xtreem88: `mkdir` is already a static method in `FileManagement`

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -209,6 +100,137 @@ export const withCioNotificationsXcodeProject: ConfigPlugin<
...
+    bundleVersion,
+    iosPath,
+    iosDeploymentTarget,
+    useFrameworks,
+  } = options;
+
+  await injectCIONotificationPodfileCode(iosPath, useFrameworks);
+
+  FileManagement.mkdir(`${iosPath}/${CIO_NOTIFICATION_TARGET_NAME}`, {
+    recursive: true,
```
Shahroz16, line 120: ${iosPath}/${CIO_NOTIFICATION_TARGET_NAME} is being used multiple times, should we make a new const  `notificationTargetDirectory` from it and use it?

package.json
```diff
@@ -40,7 +40,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": "*"
+    "customerio-reactnative": ">=1.0.1-alpha.3"
```
levibostian, line 43: The version requirement should be: 
`>= 1.0.1, < 2.0` This code as it is in this PR, we will install v2 which we don't want to do. We may need to migrate our expo plugin for major versions so we don't want to install versions automatically. 

That's why I suggest: 

```suggestion
    "customerio-reactnative": "^1.0.0"
```

src/helpers/constants/ios.ts
levibostian:
```suggestion
export const CIO_SDK_VERSION = '~> 2.0.0-alpha.1';
```

I like the idea of specifying the true syntax of the version string here in this constant. Instead of just the SDK version number, include the `~>` here, too. 

That does mean that we will need to update code like this in the code base: 

```
// before
export const CIO_PODFILE_SNIPPET = ` pod 'CustomerIO/MessagingPushAPN', '~> ${CIO_SDK_VERSION}'`;

// after
export const CIO_PODFILE_SNIPPET = ` pod 'CustomerIO/MessagingPushAPN', '${CIO_SDK_VERSION}'`;
```

levibostian: Aman, When you say, 

> Segun, is there a way to auto pull in the latest version of iOS SDK / CocoaPods ? Same for react native package version

Are you saying the latest *minor* version of the iOS and RN SDKs? My comments in this PR thus far do that. 

When I say minor version, I am saying, `>= 1.0, <2.0` so the expo plugin doesn't automatically upgrade from v1 to v2 of an SDK. The expo plugin will install v1.3 automatically, for example, but will not install v2.0 automatically which is what we want.

levibostian (changes requested):
Blocking PR because the way we install versions of RN and iOS SDKs are not consistent. My PR comments address this.

levibostian (approved):

PR #15, opened 2022-12-01
feat: update for iOS SDK v2

levibostian:
This is a start to changes needed for the new iOS v2 update. 

PR in draft mode while I ask questions.


src/helpers/constants/ios.ts
```diff
@@ -76,44 +75,3 @@ target '${CIO_NOTIFICATION_TARGET_NAME}' do
...
-
-      target.build_configurations.each do |config|
-        puts "Setting build config settings for #{target.name}"
-        config.build_settings['APPLICATION_EXTENSION_API_ONLY'] ||= 'NO'
-      end
-    end
-  end
-  react_native_post_install(installer)
-  __apply_Xcode_12_5_M1_post_install_workaround(installer)
-end`;
```
levibostian, line 119: I removed all of this code because [it's a workaround that I believe should only be suggested by technical support when a bug is encountered](https://github.com/customerio/docs/pull/800#discussion_r1021803825). By us giving this workaround to all customers, it means we may be letting bugs go undiscovered.

src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -2,10 +2,6 @@ import type { CustomerIOPluginOptionsIOS } from '../../types/cio-types';
 import {
   CIO_PODFILE_REGEX,
   CIO_PODFILE_SNIPPET,
-  CIO_PODFILE_POST_INSTALL_REGEX,
-  CIO_PODFILE_TARGET_NAMES_SNIPPET,
-  CIO_PODFILE_POST_INSTALL_SNIPPET,
-  CIO_PODFILE_POST_INSTALL_FALLBACK_SNIPPET,
```
levibostian, line 8: By removing these constants, this file no longer compiles. 

That's why this PR is currently in draft mode - I had a hard time understanding what the code in this file was doing. So, I decided to ask for help.

xtreem88 (commented):
After testing with the latest React Native SDK, `perbuild` succeeded. Here is the generated `Podfile`
```
require File.join(File.dirname(`node --print "require.resolve('expo/package.json')"`), "scripts/autolinking")
require File.join(File.dirname(`node --print "require.resolve('react-native/package.json')"`), "scripts/react_native_pods")
require File.join(File.dirname(`node --print "require.resolve('@react-native-community/cli-platform-ios/package.json')"`), "native_modules")

require 'json'
podfile_properties = JSON.parse(File.read(File.join(__dir__, 'Podfile.properties.json'))) rescue {}

platform :ios, podfile_properties['ios.deploymentTarget'] || '12.4'
install! 'cocoapods',
  :deterministic_uuids => false

target 'testiosapp' do
  use_expo_modules!
  config = use_native_modules!

  use_frameworks! :linkage => podfile_properties['ios.useFrameworks'].to_sym if podfile_properties['ios.useFrameworks']

  # Flags change depending on the env values.
  flags = get_default_flags()

  use_react_native!(
    :path => config[:reactNativePath],
    :hermes_enabled => flags[:hermes_enabled] || podfile_properties['expo.jsEngine'] == 'hermes',
    :fabric_enabled => flags[:fabric_enabled],
    # An absolute path to your application root.
    :app_path => "#{Dir.pwd}/.."
  )

  # Uncomment to opt-in to using Flipper
  # Note that if you have use_frameworks! enabled, Flipper will not work
  #
  # if !ENV['CI']
  #   use_flipper!()
  # end

  post_install do |installer|
    react_native_post_install(installer)
    __apply_Xcode_12_5_M1_post_install_workaround(installer)

    # This is necessary for Xcode 14, because it signs resource bundles by default
    # when building for devices.
    installer.target_installation_results.pod_target_installation_results
      .each do |pod_name, target_installation_result|
      target_installation_result.resource_bundle_targets.each do |resource_bundle_target|
        resource_bundle_target.build_configurations.each do |config|
          config.build_settings['CODE_SIGNING_ALLOWED'] = 'NO'
        end
      end
    end
  end

  post_integrate do |installer|
    begin
      expo_patch_react_imports!(installer)
    rescue => e
      Pod::UI.warn e
    end
  end
end

  pod 'CustomerIO/MessagingPushAPN', '~> 2.0.0-alpha.1'
end
```

xtreem88:
> @xtreem88 seeing the generated `Podfile` is super, super helpful. Thank you! It would be great if we can setup automated tests in the future to perform a snapshot test of the `Podfile`.

Yes, adding tests is one of the requirements for us to get to a stable release. I haven't figured out how we would do that yet because the focus now is on `beta` release

ssrballin:
Hi, I've been working on integrating this library as we are starting some new holiday campaigns depending on customerio. 

Was wondering if this version eliminate steps needed after prebuild using expo configs? And if so would it be possible for me to use this version somehow?

levibostian:
Marking as do not merge until Segun makes some updates to the branch/PR. 

Specifically, we need: 
1. Update the PR to using iOS SDK v2 stable instead of alpha/beta that is is currently using.
2. I need some help finishing up the PR by testing it, getting it to compile, etc. My machine is not setup for local development of the project so I couldn't do this myself.

src/helpers/native-files/ios/NotificationService.swift
```diff
@@ -9,6 +10,7 @@ public class NotificationServiceCioManager : NSObject {
 
     @objc(didReceive:withContentHandler:)
     public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        CustomerIO.initialize(siteId: "", apiKey: "", region: .US) { _ in }
```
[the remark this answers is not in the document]
levibostian: This line of code needs to be modified to use the site id, api key, and region of the customer. 

I am trying to find in the Expo plugin codebase where we modify the AppDelegate to call `CustomerIO.initialize`. This line of code should be identical to that.

src/helpers/constants/ios.ts
```diff
@@ -1,13 +1,13 @@
 export const LOCAL_PATH_TO_CIO_NSE_FILES = `node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
 export const IOS_DEPLOYMENT_TARGET = '13.0';
-export const CIO_SDK_VERSION = '~> 2.0.0-alpha.1';
+export const CIO_SDK_VERSION = '2.0.0';
```
levibostian, line 3:
```suggestion
export const CIO_SDK_VERSION = '~> 2.0.0';
```

Is there a reason the `~>` was removed?

src/helpers/constants/ios.ts
```diff
@@ -1,13 +1,13 @@
...
 export const CIO_PODFILE_REGEX = /pod 'CustomerIO\/MessagingPushAPN'/;
 export const CIO_CIO_TARGET_REGEX = /cio_target_names/;
 export const CIO_PODFILE_NOTIFICATION_REGEX = /target 'NotificationService' do/;
+export const CIO_PODFILE_POST_INSTALL_REGEX = /post_install do \|installer\|/;
```
levibostian, line 7: I don't see this used anywhere. Can we delete it?

src/helpers/native-files/ios/NotificationService.swift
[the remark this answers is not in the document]
levibostian: Oh, wait. I think that Expo customers [use the RN SDK for SDK initialization](https://customer.io/docs/sdk/react-native/getting-started/#initialize-and-configure-the-sdk). 

Does the Expo plugin ask the customer for site id, api key and region? So we can take that information into the plugin and inject it here?

src/helpers/constants/ios.ts
xtreem88: It is used here: https://github.com/customerio/customerio-expo-plugin/blob/edce68ce8cb269c3ba7f98e1b38f71af49a1b390/src/helpers/utils/injectCIOPodfileCode.ts#L22

levibostian:
@xtreem88 Is there a reason this PR was closed?

PR #18, opened 2022-12-19
fix: fixed info.plist formatting issue

xtreem88:
https://github.com/customerio/issues/issues/8857

The function that should handle replacements in `.plist` file calls `injectCodeByMultiLineRegex`, which is wrong since this function does not replace the values; it only injects a new value.
This PR adds a new function, `replaceCodeByRegex` which does the replacement correctly


matt-frizzell (approved):
Yup. All good to me.

PR #22, opened 2022-12-21
fix: minor fix with path

xtreem88:
A customer got an error related to a file path: https://secure.helpscout.net/conversation/2079296191/184737/
It is unclear what caused the error because they had been able to run prebuild before and only got an error when trying to run their app.

This PR attempts to fix this issue with the path.


matt-frizzell (commented):
This looks fine if it fixes the problem and continues to work otherwise. It does change the nature of the path we are now supplying. Were you able to reproduce the error and then test this change?

matt-frizzell (approved):
Approved with the expectation that this was tested and solved

xtreem88:
Closing because this branch will be released on `beta` and is part of this PR: https://github.com/customerio/customerio-expo-plugin/pull/33

PR #23, opened 2022-12-23
feat: rich push environment variable setup

xtreem88:
This PR resolves: https://github.com/customerio/issues/issues/8922 and also inherits changes from https://github.com/customerio/customerio-expo-plugin/pull/15

Users can provide the following details:
```
"env": {
  "siteId": "94541e1bbff594682089",
  "apiKey": "914ea21ebfa87bf771f4",
  "region": "us"
}
```
These details would be used to setup the rich push workaround


src/helpers/native-files/ios/Env.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+class Env {
+    static var customerIOSiteId: String = "{{SITE_ID}}"
```
matt-frizzell, line 4: @xtreem88 Remind me again, where do these values come in from?

matt-frizzell: I see where the replacement happens down in withNotificationXCodeProeject.ts, I'm wondering about the values.

matt-frizzell: OH WAIT. This is probably in the documentation. Does the person using the file need to fill in those placeholders first for all of this to work? (That would make sense to me)

src/helpers/native-files/ios/NotificationService.swift
```diff
@@ -9,6 +10,8 @@ public class NotificationServiceCioManager : NSObject {
 
     @objc(didReceive:withContentHandler:)
     public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        CustomerIO
+            .initialize(siteId: Env.customerIOSiteId, apiKey: Env.customerIOApiKey, region: Region.US) { config in }
```
ami-aman, line 14: ` Region.US`
Region can be US or EU. But hard coding to US could create a problem for customers having EU region.

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -260,6 +263,43 @@ const updateNseInfoPlist = (payload: {
...
+      API_KEY_RE,
+      options.pushNotification?.env?.apiKey
+    );
+  }
+
+  if (options.pushNotification?.env?.region) {
+    envFileContent = replaceCodeByRegex(
+      envFileContent,
+      REGION_RE,
+      options.pushNotification?.env?.region
```
ami-aman, line 296: Region is a string in [Env class](https://github.com/customerio/customerio-expo-plugin/blob/7c6746f2f7a31ac4a6449a5ea6eb1cad207ef9a7/src/helpers/native-files/ios/Env.swift#L6) and looks like  string is getting updated in initialisation in NotificationExtension.

I could be wrong too. 🤷

src/helpers/native-files/ios/NotificationService.swift
```diff
@@ -9,6 +10,8 @@ public class NotificationServiceCioManager : NSObject {
 
     @objc(didReceive:withContentHandler:)
     public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        CustomerIO
+            .initialize(siteId: Env.customerIOSiteId, apiKey: Env.customerIOApiKey, region: Env.customerIORegion) { config in }
```
ami-aman, line 14: I doubt if this will work. The 3rd parameter region is of `Region` type but in Env.swift file region is of type `string`. Using the value of string type here in the code above will throw an error. Did you check if this executes and creates a build ?

src/ios/withNotificationsXcodeProject.ts
ami-aman: Re-reading my comment sounds like I didn't explain well. I added an improvised comment [here](https://github.com/customerio/customerio-expo-plugin/pull/23/files#r1058471707)

src/helpers/native-files/ios/NotificationService.swift
xtreem88: I have pushed a fix for this

ami-aman (approved):

smlarkin:
> This PR resolves: https://github.com/customerio/issues/issues/8922 and also inherits changes from #15
> 
> Users can provide the following details:
> 
> ```
> "env": {
>   "siteId": "94541e1bbff594682089",
>   "apiKey": "914ea21ebfa87bf771f4",
>   "region": "us"
> }
> ```
> 
> These details would be used to setup the rich push workaround

@xtreem88 @ami-aman ... 

Can anyone tell me where I find these keys ( siteId, apiKey, region ) ?

Are they specific to the app in AppConnect or Customer.io ?

Your docs also do not indicate this or point in any direction ( see attached ) .

customerIO_Expo

Thanks in advance.

smlarkin:
@xtreem88 @ami-aman   Also, I believe your Repo & Docs are outdated since this PR is merged.  Specifically, it seems that the direction to alter the [ios/NotificationService/NotificationService.swift](https://github.com/customerio/customerio-expo-plugin#enable-rich-push-on-ios) would no longer be necessary -- Is that true ?

xtreem88:
Hello @smlarkin 
Thanks for the feedback.
To find your API credentials, see this doc: https://customer.io/docs/managing-credentials/#where-can-you-find-them
We do need to be clearer in our Expo docs about this. We will make some changes to fix that.
To answer your second question, the docs and repo are up to date, and the file still exists in our repo.
Are you having any issues with the plugin? I can help you if you need help.

PR #33, opened 2023-02-08
feat: make region case insensitive

xtreem88:
This PR resolves: https://github.com/customerio/issues/issues/9349

It modifies the logic that checks the region value to be case insensitive
Also, improves the information in the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ The `customerio-expo-plugin` supports the following configuration options. In mo
             <td><code>android.googleServicesFile</code></td>
             <td>string</td>
             <td><code>undefined</code></td>
-            <td>Set the path to the folder that contains your <code>google-services.json</code> file with a trailing slash (<code>/</code>).</td>
+            <td>Set the path to your <code>google-services.json</code> file.</td>
         </tr>
         <tr>
             <td><code>ios.pushNotification</code></td>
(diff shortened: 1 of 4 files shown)


src/ios/withNotificationsXcodeProject.ts
```diff
@@ -291,9 +291,9 @@ const updateNseEnv = (
 
   if (options.pushNotification?.env?.region) {
     let region = '';
-    if (options.pushNotification?.env?.region === 'us') {
+    if (options.pushNotification?.env?.region?.toLowerCase() === 'us') {
```
matt-frizzell, line 294: Do all our SDKs make this same decision on lower case versus upper case?

mrehan27 (approved):

README.md
```diff
@@ -136,6 +136,12 @@ The `customerio-expo-plugin` supports the following configuration options. In mo
             <td><code>false</code></td>
             <td>Enables rich push for iOS</td>
         </tr>
+        <tr>
+            <td style="white-space:nowrap;"><code>ios.pushNotification.env</code></td>
+            <td>object</td>
+            <td><code>undefined</code></td>
+            <td>Set environment variables to use for rich push workaround. This field should be filled when enabling rich push. Expected values: `siteId`: `string`,`apiKey`: `string`, `region`: `us` or `eu`</td>
```
mrehan27, line 143: Not specific to this PR, but I couldn't find any sample object for this. Should we add an example object so it is clear for customers?

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -291,9 +291,9 @@ const updateNseEnv = (
...
       region = 'Region.US';
-    } else if (options.pushNotification?.env?.region === 'eu') {
+    } else if (options.pushNotification?.env?.region?.toLowerCase() === 'eu') {
       region = 'Region.EU';
     }
```
Shahroz16, line 298: Can we maybe avoid the changing? how does this look? 

```suggestion
     const regionMap = {
  us: 'Region.US',
  eu: 'Region.EU',
};

const region = options.pushNotification?.env?.region?.toLowerCase();
const mappedRegion = regionMap[region] || '';
```

README.md
xtreem88: This is part of the docs cleanup we will be doing soon with @glosier's help

Shahroz16 (approved):

PR #37, opened 2023-02-16
fix: hot fixes

xtreem88:
This PR combined changes from these PRs: 
https://github.com/customerio/customerio-expo-plugin/pull/29
https://github.com/customerio/customerio-expo-plugin/pull/33

Both branches were combined so all changes could be tested before merging to `beta`

### DIFF
--- a/src/ios/withNotificationsXcodeProject.ts
+++ b/src/ios/withNotificationsXcodeProject.ts
@@ -299,13 +299,23 @@ const updateNseEnv = (
   }
 
   if (options.pushNotification?.env?.region) {
-    let region = '';
-    if (options.pushNotification?.env?.region === 'us') {
-      region = 'Region.US';
-    } else if (options.pushNotification?.env?.region === 'eu') {
-      region = 'Region.EU';
+    const regionMap = {
+      us: 'Region.US',
+      eu: 'Region.EU',
+    };
+    const region = options.pushNotification?.env?.region?.toLowerCase();
+    const mappedRegion = (regionMap as any)[region] || '';
+    if (!mappedRegion) {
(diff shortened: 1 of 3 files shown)


mrehan27:
@xtreem88 Links to branches aren't working because they have been deleted, I think we should update it to PR links?

mrehan27 (approved):

xtreem88:
> @xtreem88 Links to branches in PR description aren't working because they have been deleted, I think we should update it to PR links?

@mrehan27 I've updated the description to the PRs instead of the branches

src/ios/withNotificationsXcodeProject.ts
```diff
@@ -290,13 +299,23 @@ const updateNseEnv = (
...
-    if (options.pushNotification?.env?.region === 'us') {
-      region = 'Region.US';
-    } else if (options.pushNotification?.env?.region === 'eu') {
-      region = 'Region.EU';
+    const regionMap = {
+      us: 'Region.US',
+      eu: 'Region.EU',
+    };
+    const region = options.pushNotification?.env?.region?.toLowerCase();
+    const mappedRegion = (regionMap as any)[region] || '';
```
ami-aman, line 307: Wondering if the user doesn't pass value as "US" or "EU" but anything else for example, "PU", would this return an empty string or `undefined` ? Because the next condition checks that `if (!mappedRegion)` which will never be true ? Please correct me if I understood it wrong!

xtreem88: It would return an empty string `if (!mappedRegion)` checks for empty string, undefined, null... etc

PR #40, opened 2023-02-21
feat: flexible notification request handling

xtreem88:
Resolves: https://github.com/customerio/issues/issues/9479

Android apk: https://expo.dev/artifacts/eas/oH3ucFEUnHdY9vpALEA2u8.apk


src/android/withAndroidManifestUpdates.ts
```diff
@@ -0,0 +1,49 @@
...
+export const withAndroidManifestUpdates: ConfigPlugin<
+  CustomerIOPluginOptionsAndroid
+> = (configOuter) => {
+  return withAndroidManifest(configOuter, (props) => {
+    const application = props.modResults.manifest
+      .application as ManifestApplication[];
+    const customerIOMessagingpush =
+      'io.customer.messagingpush.CustomerIOFirebaseMessagingService';
+
+    if (!application[0]['service']) {
```
mrehan27, line 15: Question: What does `application[0]` refers to? As per my understanding, it gets the `application` tag from list of manifests where `0` indicates it'll get `application` from app level manifest which is highest in priority. Is this correct?

mrehan27 (approved):
Just one question, if my assumptions are correct, Android part seems good to me 👍🏻

xtreem88:
> Just one question, if my assumptions are correct, Android part seems good to me 👍🏻

Correct!

src/android/withAndroidManifestUpdates.ts
xtreem88: Yes, correct! 
This is a copy of the type definition:
```
export declare type AndroidManifest = {
    manifest: {
        $: {
            'xmlns:android': string;
            'xmlns:tools'?: string;
            package?: string;
            [key: string]: string | undefined;
        };
        permission?: ManifestPermission[];
        'uses-permission'?: ManifestUsesPermission[];
        'uses-permission-sdk-23'?: ManifestUsesPermission[];
        'uses-feature'?: ManifestUsesFeature[];
        application?: ManifestApplication[];
    };
};
```

ami-aman:
I tested the branch on a sample app and found that on `disableNotificationRegistration: true`, the plugin did not add `registerPushNotification` method call in AppDelegate's `didFinishLaunchingWithOptions` but `CIOAppPushNotificationsHandler` class in `PushService.swift` file still had `registerPushNotification` method declaration. 

This was tested using `prebuild` & `prebuild --clean` and both had same results in PushService.swift

xtreem88:
> didFinishLaunchingWithOptions

Fixed!

xtreem88:
> I tested the branch on a sample app and found that on `disableNotificationRegistration: true`, the plugin did not add `registerPushNotification` method call in AppDelegate's `didFinishLaunchingWithOptions` but `CIOAppPushNotificationsHandler` class in `PushService.swift` file still had `registerPushNotification` method declaration.
> 
> This was tested using `prebuild` & `prebuild --clean` and both had same results in PushService.swift

Fixed!

README.md
```diff
@@ -124,6 +124,12 @@ The `customerio-expo-plugin` supports the following configuration options. In mo
             <td><code>undefined</code></td>
             <td>Set the path to your <code>google-services.json</code> file.</td>
         </tr>
+        <tr>
+            <td><code>android.setHighPriorityPushHandler</code></td>
+            <td>boolean</td>
+            <td><code>undefined</code></td>
+            <td>This is optional, if you choose to use a 3rd party plugin to handle notification permissions, but want our SDK to handle the notifications, this sets `io.customer.messagingpush.CustomerIOFirebaseMessagingService` as high priority.</td>
+        </tr>
```
Shahroz16, line 132: I think we can skip the high priority part because we are putting priority as `-1`.

Shahroz16 (approved):

PR #51, opened 2023-03-06
feat: support for React Native and iOS SDK v2

xtreem88:
Resolves: https://github.com/customerio/issues/issues/8792

This adds support for iOS SDk v2 only.

## QA Instructions
- Create a new Expo project or use an existing one
- install Expo plugin on this branch
- install the latest RN SDK
- run `expo prebuild --clean`
- install older versions of RN SDK and run `expo prebuild --clean` for each version
- also set `ios.disableNotificationRegistration` to `true` and then `false` and also leave it out

## Expected Result
Prebuild should only be successful only on RN SDK with iOS SDK v2 or higher.
RN SDK with iOS SDK lower than v2 should throw an error relating to dependencies.
The code for registering push notification should only be added when `ios.disableNotificationRegistration` is set to `false`

### DIFF
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -1,6 +1,6 @@
 export const LOCAL_PATH_TO_CIO_NSE_FILES = `./node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
 export const IOS_DEPLOYMENT_TARGET = '13.0';
-export const CIO_SDK_VERSION = "'>= 2.0.0', '< 3.0'";
+export const CIO_SDK_VERSION = "'~> 2.0'";
 export const CIO_PODFILE_REGEX = /pod 'CustomerIO\/MessagingPushAPN'/;
 export const CIO_CIO_TARGET_REGEX = /cio_target_names/;
 export const CIO_PODFILE_NOTIFICATION_REGEX = /target 'NotificationService' do/;
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -1,6 +1,6 @@
 export const LOCAL_PATH_TO_CIO_NSE_FILES = `./node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
 export const IOS_DEPLOYMENT_TARGET = '13.0';
-export const CIO_SDK_VERSION = '~> 2.0.0';
+export const CIO_SDK_VERSION = "'>= 2.0.0', '< 3.0'";
 export const CIO_PODFILE_REGEX = /pod 'CustomerIO\/MessagingPushAPN'/;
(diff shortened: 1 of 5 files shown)


src/helpers/constants/ios.ts
```diff
@@ -1,6 +1,6 @@
 export const LOCAL_PATH_TO_CIO_NSE_FILES = `./node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
 export const IOS_DEPLOYMENT_TARGET = '13.0';
-export const CIO_SDK_VERSION = '~> 2.0.0';
+export const CIO_SDK_VERSION = "'>= 2.0.0', '< 3.0'";
```
Shahroz16, line 3: any particular reason we moved away from the previous syntax? because I think `~> 2.0.0` means the same thing.

xtreem88: Customers complained of the following error when we use `~> 2.0.0`
```
Command `pod install` failed.
└─ Cause: CocoaPods could not find compatible versions for pod "CustomerIOTracking":
  In Podfile:
    CustomerIO/MessagingPushAPN (~> 2.0.0) was resolved to 2.0.0, which depends on
      CustomerIOMessagingPushAPN (= 2.0.0) was resolved to 2.0.0, which depends on
        CustomerIOMessagingPush (= 2.0.0) was resolved to 2.0.0, which depends on
          CustomerIOTracking (= 2.0.0)

    customerio-reactnative (from `../node_modules/customerio-reactnative`) was resolved to 2.2.0, which depends on
      CustomerIOMessagingInApp (~> 2.1.0-beta.2) was resolved to 2.1.0, which depends on
        CustomerIOTracking (= 2.1.0)

    customerio-reactnative (from `../node_modules/customerio-reactnative`) was resolved to 2.2.0, which depends on
      CustomerIOTracking (~> 2.1.0-beta.2)

```

ami-aman: Wondering the same as Shahroz. Rest, the PR looks good to me.

ami-aman (approved):

src/helpers/constants/ios.ts
levibostian:
```suggestion
export const CIO_SDK_VERSION = '~> 2.0';
```

I think that this syntax is the same as `>= 2.0.0', '< 3.0`. 

I think this is the case based off of [the docs](https://guides.cocoapods.org/syntax/podfile.html#pod) for `~>` behavior. 

I did try `'~> 2.0'` in a cocoapods project and it did install 2.1 of the iOS SDK.

xtreem88: Using `'~> 2.0'` also works

levibostian (approved):

wyeo:
Please merge !

PR #67, opened 2023-04-28
fix: compatibility for monorepos

xtreem88:
Resolves: https://github.com/customerio/issues/issues/9943

## Approach
- Find the `customerio-expo-plugin` module file path
- Get the root directory of the `customerio-expo-plugin` module
- Navigate to helpers directory, relative to the root directory

## Testing
- clone a monorepo with an Expo app
- run `yarn link customerio-expo-plugin` on the root directory
- cd into the Expo app directory
- run `yarn add customerio/customerio-expo-plugin#expo-compatibility-for-monorepos`
- run `yarn add customerio-reactnative`
- setup `app.json` /`app.config.js`
- run `expo prebuild --clean`

### Expectation
- `prebuild` should finish without an error
- test the app to see that it builds and runs

### DIFF
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -1,4 +1,14 @@
-export const LOCAL_PATH_TO_CIO_NSE_FILES = `./node_modules/customerio-expo-plugin/src/helpers/native-files/ios`;
+const findWorkspaceRoot = require('find-workspace-root');
+const workspaceRoot = findWorkspaceRoot();
+const path = require('path');
+const nodeModulesPath = workspaceRoot
+  ? path.join(workspaceRoot, 'node_modules')
+  : './node_modules';
+
+export const LOCAL_PATH_TO_CIO_NSE_FILES = path.join(
+  nodeModulesPath,
+  'customerio-expo-plugin/src/helpers/native-files/ios'
+);
 export const IOS_DEPLOYMENT_TARGET = '13.0';
 export const CIO_SDK_VERSION = "'~> 2.0'";
 export const CIO_PODFILE_REGEX = /pod 'CustomerIO\/MessagingPushAPN'/;
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -1,13 +1,10 @@
+const finder = require('find-package-json');
 const path = require('path');
 
-// Find the path to the package
-let pluginIndexPath = require.resolve('customerio-expo-plugin');
-
-// remove index.js from the end of the pluginIndexPath
-pluginIndexPath = path.dirname(pluginIndexPath);
-
+const f = finder(__dirname);
+let pluginPackageRoot = f.next().filename;
 // This is the path to the root of the customerio-expo-plugin package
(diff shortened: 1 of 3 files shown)


src/helpers/constants/ios.ts
```diff
@@ -1,4 +1,18 @@
...
+const path = require('path');
+
+// Find the path to the package
+let pluginIndexPath = require.resolve('customerio-expo-plugin');
+
+// remove index.js from the end of the pluginIndexPath
+pluginIndexPath = path.dirname(pluginIndexPath);
+
+// This is the path to the root of the customerio-expo-plugin package
+const pluginPackageRoot = path.join(pluginIndexPath, '..', '..');
```
levibostian, line 10: If `index.js` in the `customerio-expo-plugin` module ever moves, will this path of `../../` still be valid? 

Is there an alternative way to find the root directory of `node_modules/customerio-expo-plugin` that works no matter if the file structure of this module changes?

xtreem88: index.js is the module's entry point; unless we change the structure of the project, it wont change location

levibostian: Is there an alternative way to find the root directory of `node_modules/customerio-expo-plugin` that works even if the file structure of this module changes?

xtreem88: I'm not sure if there's another way.
We basically have to find a reference point that we are sure of, and then from that reference point, we can navigate to where we want. The surest point of reference is the module entry file, if for any reason we move that file, we have to change other configurations for the plugin to work, so there should be no benefit to moving that file in the first place

levibostian: What if we use the `package.json` file as a reference point instead? Since `package.json` should always be in the root directory: `node_modules/customerio-expo-plugin`. 

[Here is a small package](https://github.com/sindresorhus/pkg-dir) that seems like it would help us solve this problem.

xtreem88: This package is an ES Module, and using it with our plugin would require using a dynamic import which can be problematic. changing the structure to use this would mean we have to use an ES Module also and we would be [violating a rule of creating config plugins](https://docs.expo.dev/config-plugins/plugins-and-mods/#import-a-plugin)
`No import/export keywords, use module.exports in the shipped plugin file.`

xtreem88: I'll look for a plugin that can do something similar or just use their approach

xtreem88: I found a CommonJS [package](https://www.npmjs.com/package/find-package-json) that does the same job, I tested, and it seems good enough.

levibostian (approved):

sergey-king:
Hey guys, any updates on when this will be merged & released?

sergey-king:
Hey guys, any updates on when this will be merged & released?
@levibostian @xtreem88

dannyhw:
you can use node directly to resolve the location, this is the patch I use

```
const path = require("path");
const LOCAL_PATH_TO_CIO_NSE_FILES = path.resolve(require.resolve('customerio-expo-plugin/postinstall'), "../helpers/native-files/ios")
```

Issue #72, opened 2023-05-15
[Bug]: Customer.IO overwrites project file changes from other plugins


euc-callum:
Hi,

We're attempting to utilise `expo-notifications` alongside `customerio/customerio-expo-plugin` for richpush notifications on iOS.

Whilst investigating writing our own Expo plugin to handle registering the entitlement file in the project file (a task on the Customer.IO roadmap with no timeline), we discovered this is impossible due to the manner in which Customer.IO is writing project files changes. We also discovered that it shouldn't be necessary as `expo-notifications` is already configuring it, but Customer.IO is then overwriting those changes.

**I believe this would also impact any other plugin that modifies the iOS project files that exists in a project utilising `customerio/customerio-expo-plugin`, potentially overwriting any changes they make, which adds a lot of risk to utilising `customerio/customerio-expo-plugin`.**

We believe the problems to be as follows:

Firstly, the plugin does not await asynchronous actions within it, and thus rather than running in sequence with other plugins that are modifying the project file, it will instead overlap them. This will result in race conditions where the behaviour of plugins will vary based on the timing of when the project file is read by a given plugin.

See [here](https://github.com/customerio/customerio-expo-plugin/blob/64ba9326f601c1b5bc5bd7a03c4bea57494ae8ee/src/ios/withNotificationsXcodeProject.ts#L31) for an example that should be promisified. I've included what this would look like below:

```
await new Promise((resolve, reject) => {
  xcodeProject.parse(async function (err: Error) {
    if (err) {
      reject(`Error parsing iOS project: ${JSON.stringify(err)}`);
    }

    if (options.pushNotification) {
      await addPushNotificationFile(options, xcodeProject);
    }

    if (options.pushNotification?.useRichPush) {
      await addRichPushXcodeProj(options, xcodeProject);
    }

    FileManagement.writeFile(projPath, xcodeProject.writeSync());
    resolve();
  });
});
```

This change alone will not resolve things however, and will actually result in Customer.IO not working, as the changes that Customer.IO has then written to the project file will then be overwritten by Expo. See [here](https://docs.expo.dev/config-plugins/plugins-and-mods/#mod-plugins) for information on how Expo suggests to modify files via returning `modResults` rather than writing to files.

I believe Customer.IO's Expo plugin code needs modifying to utilise `withXcodeProject` and other similar variants rather than directly modifying files in order for it to work correctly, alongside ensuring it only returns once all asynchronous actions have completed.

[...]

PR #73, opened 2023-05-15
fix: resolve issues with saving the project file async

euc-callum:
See https://github.com/customerio/customerio-expo-plugin/issues/72.

This change ensures that the Expo plugin returns any changes that it makes to the project file rather than saving these asynchronously (and not awaiting the result).

### DIFF
--- a/src/types/cio-types.ts
+++ b/src/types/cio-types.ts
@@ -25,7 +25,6 @@ export type CustomerIOPluginOptionsIOS = {
       region: string;
     };
   };
-  xcodeProject: any;
 };
 
 export type CustomerIOPluginOptionsAndroid = {
--- a/src/types/cio-types.ts
+++ b/src/types/cio-types.ts
@@ -25,6 +25,7 @@ export type CustomerIOPluginOptionsIOS = {
       region: string;
     };
   };
+  xcodeProject: any;
 };
 
 export type CustomerIOPluginOptionsAndroid = {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "customerio-expo-plugin",
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "customerio-expo-plugin",
-      "version": "1.0.0-beta.6",
+      "version": "1.0.0-beta.8",
       "hasInstallScript": true,
       "license": "MIT",
       "devDependencies": {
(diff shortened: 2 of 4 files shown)


src/types/cio-types.ts
```diff
@@ -25,6 +25,7 @@ export type CustomerIOPluginOptionsIOS = {
       region: string;
     };
   };
+  xcodeProject: any;
```
levibostian, line 28: I am OK with making the Expo plugin more flexible by providing an xcode project path. 

This PR seems to make this a required step with no default value. I would prefer if there was a default value such as what the plugin currently uses: 


```ts
  const projPath = `${iosPath}/${appName}.xcodeproj/project.pbxproj`;
```

but allow overriding this value.

levibostian (commented):
Thanks for the PR! Greatly appreciated!

src/types/cio-types.ts
euc-callum: Hmm, fair point, I didn't realise this was more than just an internal type.

My aim is not to allow this to be configured, but instead to default to utilising the means Expo provides for modifying projects.

To resolve this, I'm no longer extending the options object and instead passing it directly.

euc-callum:
Is this likely to make it in @levibostian?

sergey-king:
I also need this. Subscribed

euc-callum:
Tagging another maintainer, apologies @xtreem88 but you seem active in here.

euc-callum:
Actually looks like you're trying to do something similar @xtreem88, might be a duplication of work here?

https://github.com/customerio/customerio-expo-plugin/pull/76

xtreem88:
@euc-callum Thanks for the PR; I've been away for some days; I was moving. 
I created a new [PR](https://github.com/customerio/customerio-expo-plugin/pull/76) based on your [recommendation here](https://github.com/customerio/customerio-expo-plugin/issues/72), The PR builds on top of changes you made here and also has [Appdelegate Subscriber](https://github.com/customerio/customerio-expo-plugin/pull/74). I merged both changes to be tested and QA as one release.

euc-callum:
Thanks for the heads up @xtreem88, I've commented on your pull request because I believe it may still have the issue I have resolved in this pull request.

Can I advise merging in this pull request as it is known to work (sample size of 1 - we're running this change in production now)?

sergey-king:
@xtreem88 when can we expect a new release including this fix?

StephMoreno:
Hi @sergey-king, I'm Stephanie, a Product Manager here. I'm sorry to hear about the challenges and blockers this has created for you. Regarding the extension of your free trial, can you please send us an email about this to **win@customer.io**? Our support team will be able to assist you there regarding your trial.

PR #74, opened 2023-05-16
used subscriber

xtreem88:


src/helpers/native-files/ios/PushService.swift
```diff
@@ -3,28 +3,22 @@ import CioMessagingPushAPN
...
     MessagingPush.shared.application(application, didRegisterForRemoteNotificationsWithDeviceToken: deviceToken)
   }
 
-  @objc(application:error:)
   public func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
     MessagingPush.shared.application(application, didFailToRegisterForRemoteNotificationsWithError: error)
   }
 
-  @objc(userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:)
   public func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
```
levibostian, line 20: [Looking at the callback functions for `ExpoAppDelegate`](https://github.com/expo/expo/blob/main/packages/expo-modules-core/ios/AppDelegates/ExpoAppDelegate.swift), I don't see any callbacks for user notification center? 

Will this function get called from this change?

zdnk: I forked this repo and was exploring this as well and hit the wall with this. I am not sure how it should be actually done, because it is as you are saying. ExpoAppDelegate only swizzles the basic AppDelegate methods, but not the other ones from UserNotifications etc. I believe at least, I havent tested it.

PR #76, opened 2023-05-19
fix: xcode race condition fix

xtreem88:
Resolves https://github.com/customerio/customerio-expo-plugin/issues/72 
Partially resolves: https://github.com/customerio/issues/issues/10195
Build on the changes in https://github.com/customerio/customerio-expo-plugin/pull/73

## Changes
- Used `withXcodeProject` to modify `project.pbxproj` instead of `xcode` package

## Testing instruction
- Install the plugin
- run rebuild and then build (on local and EAS) and run
- Make sure push rich push and all other features work as extected


src/ios/withNotificationsXcodeProject.ts
```diff
@@ -28,10 +39,8 @@ const addNotificationServiceExtension = async (
 
   const xcodeProject = xcode.project(projPath);
 
-  xcodeProject.parse(async function (err: Error) {
-    if (err) {
-      throw new Error(`Error parsing iOS project: ${JSON.stringify(err)}`);
-    }
+  try {
+    await parseXcodeProject(xcodeProject);
```
euc-callum, line 43: I'm curious why the `xcodeProject` needs to be parsed or retrieved from a path here?

Would it be simpler to use `modResults` from `withXCodeProject` which is specifically designed for this?

https://github.com/customerio/customerio-expo-plugin/pull/73/files#diff-3448f189b313e51968629b1f18deaa423912dfd5039f11eefe2480213d4e5661R84

I worry you'll pull a version of the xcodeProject from disk that is not equivalent to the one that is being passed in by Expo, and thus not include changes that have been provided by another plugin earlier in the chain.

levibostian (approved):

euc-callum (approved):
🙌

sergey-king:
any updates on when this will get merged?

sergey-king (approved):

PR #81, opened 2023-06-14
fix!: auto-update native SDK and easier rich push install

levibostian:
Makes the Expo plugin compatible with upcoming RN SDK version 3.0 introduced by [this pr](https://github.com/customerio/customerio-reactnative/pull/149). 

Part of ticket: https://github.com/customerio/issues/issues/10356

# Breaking change 

This PR is not backwards compatible with previous Expo plugin versions and will require customers call `expo prebuild --clean` to introduce these changes to their code. 

This PR introduces comment headers in the Podfile:
```ruby
target 'testiosapp' do
  ...
# --- CustomerIO Host App START ---
  ...
# --- CustomerIO Host App END ---
  ...
end 
```

as an idea to no longer need to ask customers to run `--clean` but they can run just `prebuild` and our plugin will make modifications to the existing `Podfile`. 

# Dependencies 

* Deploy RN SDK version 3.0 https://github.com/customerio/customerio-reactnative/pull/149

# Generated `Podfile` after prebuild: 

```ruby
require File.join(File.dirname(`node --print 
"require.resolve('expo/package.json')"`), "scripts/autolinking")
require File.join(File.dirname(`node --print 
"require.resolve('react-native/package.json')"`), "scripts/react_native_pods")
require File.join(File.dirname(`node --print 
"require.resolve('@react-native-community/cli-platform-ios/package.json')"`), 
"native_modules")

require 'json'
podfile_properties = JSON.parse(File.read(File.join(__dir__, 
'Podfile.properties.json'))) rescue {}

ENV['RCT_NEW_ARCH_ENABLED'] = podfile_properties['newArchEnabled'] == 'true' ? 
'1' : '0'
ENV['EX_DEV_CLIENT_NETWORK_INSPECTOR'] = '1' if 
podfile_properties['EX_DEV_CLIENT_NETWORK_INSPECTOR'] == 'true'

platform :ios, podfile_properties['ios.deploymentTarget'] || '13.0'
install! 'cocoapods',
  :deterministic_uuids => false

prepare_react_native_project!

# If you are using a `react-native-flipper` your iOS build will fail when 
`NO_FLIPPER=1` is set.
# because `react-native-flipper` depends on (FlipperKit,...), which will be 
excluded. To fix this,
# you can also exclude `react-native-flipper` in `react-native.config.js`
#
# ```js
# module.exports = {
#   dependencies: {
#     ...(process.env.NO_FLIPPER ? { 'react-native-flipper': { platforms: { ios: 
null } } } : {}),
#   }
# }
# ```
flipper_config = FlipperConfiguration.disabled
if ENV['NO_FLIPPER'] == '1' then
  # Explicitly disabled through environment variables
  flipper_config = FlipperConfiguration.disabled
elsif podfile_properties.key?('ios.flipper') then
  # Configure Flipper in Podfile.properties.json
  if podfile_properties['ios.flipper'] == 'true' then
    flipper_config = FlipperConfiguration.enabled(["Debug", "Release"])
  elsif podfile_properties['ios.flipper'] != 'false' then
    flipper_config = FlipperConfiguration.enabled(["Debug", "Release"], { 
'Flipper' => podfile_properties['ios.flipper'] })
  end
end

target 'testiosapp' do
  use_expo_modules!
  config = use_native_modules!

  use_frameworks! :linkage => podfile_properties['ios.useFrameworks'].to_sym if 
podfile_properties['ios.useFrameworks']
  use_frameworks! :linkage => ENV['USE_FRAMEWORKS'].to_sym if 
ENV['USE_FRAMEWORKS']

  # Flags change depending on the env values.
  flags = get_default_flags()

  use_react_native!(
    :path => config[:reactNativePath],
    :hermes_enabled => podfile_properties['expo.jsEngine'] == nil || 
podfile_properties['expo.jsEngine'] == 'hermes',
    :fabric_enabled => flags[:fabric_enabled],
    # An absolute path to your application root.
    :app_path => "#{Pod::Config.instance.installation_root}/..",
    # Note that if you have use_frameworks! enabled, Flipper will not work if 
enabled
    :flipper_configuration => flipper_config
  )
# --- CustomerIO Host App START ---
  pod 'customerio-reactnative/apn', :path => 
'../node_modules/customerio-reactnative'
# --- CustomerIO Host App END ---

  post_install do |installer|
    react_native_post_install(
      installer,
      config[:reactNativePath],
      # Set `mac_catalyst_enabled` to `true` in order to apply patches
      # necessary for Mac Catalyst builds
      :mac_catalyst_enabled => false
    )
    __apply_Xcode_12_5_M1_post_install_workaround(installer)

    # This is necessary for Xcode 14, because it signs resource bundles by 
default
    # when building for devices.
    installer.target_installation_results.pod_target_installation_results
      .each do |pod_name, target_installation_result|
      target_installation_result.resource_bundle_targets.each do 
|resource_bundle_target|
        resource_bundle_target.build_configurations.each do |config|
          config.build_settings['CODE_SIGNING_ALLOWED'] = 'NO'
        end
      end
    end
  end

  post_integrate do |installer|
    begin
      expo_patch_react_imports!(installer)
    rescue => e
      Pod::UI.warn e
    end
  end
end
# --- CustomerIO Notification START ---
  use_frameworks! :linkage => :static
  pod 'customerio-reactnative-richpush/apn', :path => 
'../node_modules/customerio-reactnative'
end
# --- CustomerIO Notification END ---
 ```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "xcode": "^3.0.1"
       },
       "peerDependencies": {
-        "customerio-reactnative": "https://github.com/customerio/customerio-reactnative/tarball/levi/native-version"
+        "customerio-reactnative": "^3.0.0"
       }
     },
     "node_modules/@ampproject/remapping": {
@@ -859,7 +859,6 @@
       "version": "7.18.6",
       "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-numeric-separator/-/plugin-proposal-numeric-separator-7.18.6.tgz",
       "integrity": "sha512-ozlZFogPqoLm8WBr5Z8UckIoE4YQ5KESVcNudyXOR8uqIkliTEgJ3RoketfG6pmzLdeZF0H/wjE9/cCEitBl7Q==",
-      "dev": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.18.6",
         "@babel/plugin-syntax-numeric-separator": "^7.10.4"
@@ -872,15 +871,15 @@
       }
     },
     "node_modules/@babel/plugin-proposal-object-rest-spread": {
-      "version": "7.18.9",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-object-rest-spread/-/plugin-proposal-object-rest-spread-7.18.9.tgz",
-      "integrity": "sha512-kDDHQ5rflIeY5xl69CEqGEZ0KY369ehsCIEbTGb4siHG5BE9sga/T0r0OUwyZNLMmZE79E1kbsqAjwFCW4ds6Q==",
+      "version": "7.20.7",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-object-rest-spread/-/plugin-proposal-object-rest-spread-7.20.7.tgz",
+      "integrity": "sha512-d2S98yCiLxDVmBmE8UjGcfPvNEUbA1U5q5WxaWFUGRzJSVAZqm5W6MbPct0jxnegUZ0niLeNX+IOzEs7wYg9Dg==",
       "dependencies": {
-        "@babel/compat-data": "^7.18.8",
-        "@babel/helper-compilation-targets": "^7.18.9",
-        "@babel/helper-plugin-utils": "^7.18.9",
+        "@babel/compat-data": "^7.20.5",
+        "@babel/helper-compilation-targets": "^7.20.7",
+        "@babel/helper-plugin-utils": "^7.20.2",
         "@babel/plugin-syntax-object-rest-spread": "^7.8.3",
-        "@babel/plugin-transform-parameters": "^7.18.8"
+        "@babel/plugin-transform-parameters": "^7.20.7"
       },
       "engines": {
         "node": ">=6.9.0"
@@ -905,12 +904,12 @@
       }
     },
     "node_modules/@babel/plugin-proposal-optional-chaining": {
-      "version": "7.18.9",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-optional-chaining/-/plugin-proposal-optional-chaining-7.18.9.tgz",
-      "integrity": "sha512-v5nwt4IqBXihxGsW2QmCWMDS3B3bzGIk/EQVZz2ei7f3NJl8NzAJVvUmpDW5q1CRNY+Beb/k58UAH1Km1N411w==",
+      "version": "7.21.0",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-optional-chaining/-/plugin-proposal-optional-chaining-7.21.0.tgz",
+      "integrity": "sha512-p4zeefM72gpmEe2fkUr/OnOXpWEf8nAgk7ZYVqqfFiyIG7oFfVZcCrU64hWn5xp4tQ9LkV4bTIa5rD0KANpKNA==",
       "dependencies": {
-        "@babel/helper-plugin-utils": "^7.18.9",
-        "@babel/helper-skip-transparent-expression-wrappers": "^7.18.9",
+        "@babel/helper-plugin-utils": "^7.20.2",
+        "@babel/helper-skip-transparent-expression-wrappers": "^7.20.0",
         "@babel/plugin-syntax-optional-chaining": "^7.8.3"
       },
       "engines": {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -40,7 +40,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": "^2.0.0"
+    "customerio-reactnative": "https://github.com/customerio/customerio-reactnative/tarball/levi/native-version"
```
levibostian, line 43:
```suggestion
    "customerio-reactnative": "^3.0.0"
```

When [this PR](https://github.com/customerio/customerio-reactnative/pull/149) gets deployed, we will need to update this line of code.

levibostian: Also, we will need to run `npm install` to update the lock file.

src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -1,38 +1,38 @@
...
 
-  if (!targetMatch) {
+  if (!matches) {
     const lines = podfile.split('\n');
+    // We need to decide what line of code in the Podfile to insert our native code.
+    // The "post_install" line is always present in an Expo project Podfile so it's reliable.
+    // Find that line in the Podfile and then we will insert our code above that line.
     const index = lines.findIndex((line) =>
-      CIO_PODFILE_POST_INSTALL_REGEX.test(line)
+      /post_install do \|installer\|/.test(line)
```
xtreem88, line 19: Is there a reason you moved the regex here instead of in the constants file?

src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -1,38 +1,38 @@
...
     );
     let content: string[] = lines;
     if (index > -1) {
+      // Constructs the new Podfile contents that we will write to the file system.
       content = [
         ...lines.slice(0, index - 1),
-        !matches ? CIO_PODFILE_SNIPPET : '',
+        hostAppTargetBlockStart,
+        hostAppTargetBlockContents,
+        hostAppTargetBlockEnd,
```
xtreem88, line 28: There are some helper functions that can help here; I think [injectCodeByMultiLineRegex](https://github.com/customerio/customerio-expo-plugin/blob/beta/src/helpers/utils/codeInjection.ts#L17) should do the trick, if not you can create a new function that can be reused

src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -1,38 +1,38 @@
...
         ...lines.slice(index - 1),
       ];
     }
 
     FileManagement.write(filename, content.join('\n'));
   } else {
-    console.log('Customerio Podfile snippets already exists. Skipping...');
+    console.log('CustomerIO Podfile snippets already exists. Skipping...');
```
xtreem88, line 35: 👍

src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -42,13 +42,19 @@ export async function injectCIONotificationPodfileCode(
...
   const filename = `${iosPath}/Podfile`;
   const podfile = await FileManagement.read(filename);
-  const matches = podfile.match(CIO_PODFILE_NOTIFICATION_REGEX);
+
+  const blockStart = '# --- CustomerIO Notification START ---';
+  const blockContent = `target 'NotificationService' do
+  ${useFrameworks === 'static' ? 'use_frameworks! :linkage => :static' : ''}
+  pod 'customerio-reactnative-richpush/apn', :path => '../node_modules/customerio-reactnative'
+end`;
+  const blockEnd = '# --- CustomerIO Notification END ---';
```
xtreem88, line 51: I think the snippets should be inside a `constants` module to separate the content from the logic. Making changes to the content would also be easier, and it should make the code more readable.

xtreem88 (approved):
LGTM! Just a few suggestions

src/helpers/utils/injectCIOPodfileCode.ts
xtreem88: This comment also applies to the constants above

src/helpers/utils/injectCIOPodfileCode.ts
levibostian: It sounds like this could be a preference. I purposely put the content inside of this file instead of the constants file because I think this method is more readable and maintainable. 

I like using constants when there is a string that is re-usable across many places in a project. Since the content in this file only is relevant to `Podfile` modifications, I like the idea of all content being encapsulated inside of this file where it's only relevant. 

During development of this feature, I found myself bouncing back and forth between this file and constants file and it took longer to read the code and understand the logic going on. When I moved the content into the file, I was able to read the code and understand it's logic much faster. 

If you still prefer keeping all strings in constants file, I am open to a 3rd opinion on the team and I am happy to make edits to this PR.

src/helpers/utils/injectCIOPodfileCode.ts
levibostian: Pushed up a commit with this refactor.

xtreem88 (approved):

PR #82, opened 2023-06-15
fix: modified regex for Expo 48 and added UNUserNotificationCenter snippet

xtreem88:
Resolves: https://github.com/customerio/issues/issues/10485

Details about bug and changes are on the main issue

QA instructions: https://github.com/customerio/issues/issues/10485#issuecomment-1593511370

### DIFF
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -20,7 +20,7 @@ export const GROUP_IDENTIFIER_TEMPLATE_REGEX = /{{GROUP_IDENTIFIER}}/gm;
 export const BUNDLE_SHORT_VERSION_TEMPLATE_REGEX = /{{BUNDLE_SHORT_VERSION}}/gm;
 export const BUNDLE_VERSION_TEMPLATE_REGEX = /{{BUNDLE_VERSION}}/gm;
 export const CIO_DIDFINISHLAUNCHINGMETHOD_REGEX =
-  /(- \(BOOL\)application:\(UIApplication \*\)application didFinishLaunchingWithOptions:\(NSDictionary \*\)launchOptions(\s|\n)*?\{)((.|\n)*)\[super(\s)application:application(\s)didFinishLaunchingWithOptions:launchOptions\];/;
+  /.*\[super(\s)application:application(\s)didFinishLaunchingWithOptions:launchOptions\];/;
 
 export const CIO_DIDFAILTOREGISTERFORREMOTENOTIFICATIONSWITHERROR_REGEX =
   /return \[super application:application didFailToRegisterForRemoteNotificationsWithError:error\];/;
@@ -63,6 +63,11 @@ export const CIO_DIDREGISTERFORREMOTENOTIFICATIONSWITHDEVICETOKEN_SNIPPET = `
 export const CIO_CONFIGURECIOSDKPUSHNOTIFICATION_SNIPPET = `
   // Register for push notifications
   [pnHandlerObj registerPushNotification:self];
+`;
+
+export const CIO_CONFIGURECIOSDKUSERNOTIFICATIONCENTER_SNIPPET = `
+  UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+  center.delegate = self;
 `;
 
 // Enable push handling - notification response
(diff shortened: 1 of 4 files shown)


levibostian (approved):

mrehan27 (approved):

src/ios/withAppDelegateModifications.ts
```diff
@@ -111,10 +122,23 @@ const addAdditionalMethodsForPushNotifications = (stringContents: string) => {
 };
 
 const addAppdelegateHeaderModification = (stringContents: string) => {
-  stringContents = injectCodeByMultiLineRegexAndReplaceLine(
-    stringContents,
+  // Add UNUserNotificationCenterDelegate if needed
+  stringContents = stringContents.replace(
     CIO_APPDELEGATEHEADER_REGEX,
-    CIO_APPDELEGATEHEADER_SNIPPET
+    (match, p1, _p2, p3) => {
```
ami-aman, line 128: I am not sure if this import statements would be added to `AppDelegate.m` or `AppDelegate.h` file. It should be added to .h file.

xtreem88: It is added to `AppDelegate.h`

src/helpers/constants/ios.ts
```diff
@@ -35,16 +35,11 @@ export const CIO_APPDELEGATEDECLARATION_REGEX =
   /@implementation AppDelegate(.|\n)/;
 
 export const CIO_APPDELEGATEHEADER_REGEX =
-  /@interface AppDelegate : EXAppDelegateWrapper <RCTBridgeDelegate>/;
+  /(@interface AppDelegate\s:\sEXAppDelegateWrapper\s?)(<([^>]+)>)?/;
```
levibostian, line 38:
```suggestion
  /(@interface AppDelegate\s*:\s*EXAppDelegateWrapper\s*)(<([^>]+)>)?/;
```

There might be more or less whitespace characters so this suggestion could help prevent errors against that.

src/ios/withAppDelegateModifications.ts
levibostian: The comments in the function are great. 

Could you use more specific variable names instead of `match, p1, _p2, p3`? I still don't quite understand this code with those names.

levibostian (approved):

levibostian (changes requested):
I am going to ask a couple questions in Slack. I suggest holding off on PR until you see conversation.

src/ios/withAppDelegateModifications.ts
```diff
@@ -111,10 +124,23 @@ const addAdditionalMethodsForPushNotifications = (stringContents: string) => {
...
+        return match;
+      } else if (existingDelegates) {
+        // Other delegates exist, append ours
+        return `${CIO_APPDELEGATEHEADER_IMPORT_SNIPPET}
+${interfaceDeclaration}<${existingDelegates}, ${CIO_APPDELEGATEHEADER_USER_NOTIFICATION_CENTER_SNIPPET}>`;
+      } else {
+        // No delegates exist, add ours
+        return `${CIO_APPDELEGATEHEADER_IMPORT_SNIPPET}
+${interfaceDeclaration.trim()} <${CIO_APPDELEGATEHEADER_USER_NOTIFICATION_CENTER_SNIPPET}>`;
+      }
```
xtreem88, line 142: I am prioritizing maintainability over readability here because the snippets are basically `magic strings` I think it makes sense to have them all in one location where they can easily be changed or edited if we have to

levibostian (approved):

src/ios/withAppDelegateModifications.ts
```diff
@@ -151,6 +179,12 @@ export const withAppDelegateModifications: ConfigPlugin<
       ) {
         stringContents = addNotificationConfiguration(stringContents);
       }
+      if (
+        props.handleNotificationClick !== undefined &&
+        props.handleNotificationClick === true
```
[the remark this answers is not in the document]
ami-aman: I do not think there is any direct way of assigning default values we are following in expo plugin. In case of a boolean, we can handle only one condition, either `true` or `false`. If the value of `handleNotificationClick` is not received then the plugin will not add the code otherwise it will, if `true`.

PR #88, opened 2023-07-01
Aman/initialise sdk ios native

xtreem88:


ami-aman:
This PR is not ready yet hence closing it.

PR #93, opened 2023-07-17
fix: handling deeplink in killed state

ami-aman:
Fixes handling deep links in killed state.

Refers : https://github.com/customerio/issues/issues/10367

### DIFF
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -38,6 +38,7 @@ export const CIO_RCTBRIDGE_DEEPLINK_MODIFIEDOPTIONS_REGEX =
 export const CIO_LAUNCHOPTIONS_DEEPLINK_MODIFIEDOPTIONS_REGEX = 
 /^\s*return\s\[\s*super\s*application:\s*application\s*didFinishLaunchingWithOptions\s*:\s*launchOptions\s*\];/gm;
 
+export const CIO_DEEPLINK_COMMENT_REGEX = /\sDeep link workaround for app killed state start/gm;
 export const DEFAULT_BUNDLE_VERSION = '1';
 export const DEFAULT_BUNDLE_SHORT_VERSION = '1.0';
 export const CIO_TARGET_NAME = 'CustomerIOSDK';
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -1,6 +1,5 @@
 const finder = require('find-package-json');
 const path = require('path');
-const Constants = require("expo-constants");
 
 const f = finder(__dirname);
 let pluginPackageRoot = f.next().filename;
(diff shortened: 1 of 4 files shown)


src/helpers/constants/ios.ts
```diff
@@ -41,6 +45,10 @@ export const CIO_APPDELEGATEHEADER_USER_NOTIFICATION_CENTER_SNIPPET = 'UNUserNot
 export const CIO_PUSHNOTIFICATIONHANDLERDECLARATION_SNIPPET = `
 CIOAppPushNotificationsHandler* pnHandlerObj = [[CIOAppPushNotificationsHandler alloc] init];
 `;
+export const CIO_RCTBRIDGE_DEEPLINK_MODIFIEDOPTIONS_SNIPPET = `
+RCTBridge *bridge = [self.reactDelegate createBridgeWithDelegate:self launchOptions:modifiedLaunchOptions];
+`;
```
levibostian, line 50: By our Expo plugin modifying the launch options, we could make our Expo plugin become incompatible with other Expo plugins that might require `launchOptions`. 

Also, from my understanding of `RCTBridge`, the API for this code has changed a lot with different recent versions of RN. Therefore, I am not sure if this code here will be backwards compatible with multiple versions of Expo.

levibostian (changes requested):

src/ios/withAppDelegateModifications.ts
```diff
@@ -185,6 +199,15 @@ export const withAppDelegateModifications: ConfigPlugin<
       ) {
         stringContents = addUserNotificationCenterConfiguration(stringContents);
       }
+
+      if (
+        props.handleDeeplinkInKilledState !== undefined &&
+        props.handleDeeplinkInKilledState == true
+      ) {
+        stringContents = addHandleDeeplinkInKilledStateConfiguration(stringContents)
+        stringContents = replaceCodeByRegex(stringContents, CIO_RCTBRIDGE_DEEPLINK_MODIFIEDOPTIONS_REGEX, CIO_RCTBRIDGE_DEEPLINK_MODIFIEDOPTIONS_SNIPPET);
```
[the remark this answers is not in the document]
ami-aman: Done ✅

src/helpers/constants/ios.ts
levibostian: I have been thinking about this and unfortunately, I do not have a suggestion to prevent the use of `launchOptions` or `RCTBridge`. 

I am mostly concerned with the use of `RCTBridge` and trying to make this code compatible with all versions of RN that our SDK supports. I know that you're testing against that, so I'll resolve this 👍

ami-aman: Thanks Levi ! I have added a fix to handle compatibility for `launchOptions` or `RCTBridge`. With this fix, the code is compatible with all existing versions of Expo (>=45 & <=48) and react native !

src/ios/withAppDelegateModifications.ts
```diff
@@ -185,6 +199,15 @@ export const withAppDelegateModifications: ConfigPlugin<
       ) {
         stringContents = addUserNotificationCenterConfiguration(stringContents);
       }
+
+      if (
+        props.handleDeeplinkInKilledState !== undefined &&
+        props.handleDeeplinkInKilledState == true
+      ) {
```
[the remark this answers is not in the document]
ami-aman: Thanks for the suggestion, Segun ! Please check the latest commit for the fix.

src/ios/withAppDelegateModifications.ts
```diff
@@ -185,6 +227,14 @@ export const withAppDelegateModifications: ConfigPlugin<
       ) {
         stringContents = addUserNotificationCenterConfiguration(stringContents);
       }
+
+      if (
+        props.handleDeeplinkInKilledState !== undefined &&
+        props.handleDeeplinkInKilledState === true
+      ) {
```
mrehan27, line 234: Isn't the following enough?

```
if (props.handleDeeplinkInKilledState === true)
```

As `undefined === true` will eventually lead to `false`

Shahroz16 (approved):

mrehan27 (approved):
LGTM

src/ios/withAppDelegateModifications.ts
ami-aman: This has been added to avoid the edge case because we do not have any direct default values of these properties.

PR #95, opened 2023-08-23
fix: Swift Bridging Headers were being added before the AppDelegate.h import

larrysingh:
This PR resolves an issue where the Swift Bridging Headers were being added before the AppDelegate.h import in the AppDelegate.mm file. The incorrect ordering of the imports causes issues when adding a Widget Extension target with a Configuration Intent to the app.


Ahmed-Ali:
First off, I apologies this PR took that long without any feedback on our end.
Second, as our expo plugin will go under some rework soon and this PR seems to be outdated, I will go ahead and close it.

PR #97, opened 2023-09-11
ci: automatically update github actions via dependabot

xtreem88:
## Changes:

- Added Dependabot configuration to monitor GitHub Actions.
- Set to check updates daily and auto-create PRs for any found updates.

## Benefits:

- Keeps our actions updated and secure.
- Automates maintenance, reducing manual effort.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "chore"
+      include: "scope"
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -3,8 +3,9 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "hourly"
-    target-branch: "cleaning/dependabot"
+      interval: "weekly"
+    reviewers:
+      - "customerio/mobile"
     commit-message:
-      prefix: "chore:"
+      prefix: "chore"
       include: "scope"
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,3 +1,5 @@
+# Dependabot helps update dependencies to keep them up-to-date. 
+# Configuration docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
```
levibostian, line 1:
```suggestion
# Dependabot helps update dependencies to keep them up-to-date. 
# Configuration docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
version: 2
```

Would be nice to have a link to the config file docs.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "hourly"
```
levibostian, line 6:
```suggestion
      interval: "weekly"
```

I don't think [hourly is an option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval). 

How about weekly? In the past, I have found daily to be a bit overwhelming where dependabot makes a lot of PRs and it's annoying to keep up with.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    target-branch: "cleaning/dependabot"
```
levibostian, line 7:
```suggestion
```

`target-branch` is the branch where code exists that should be updated. For this repo, that's `beta`. Dependabot uses the repo's default branch as the default value so we shouldn't need to specify anything.

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    commit-message:
+      prefix: "chore:"
```
levibostian, line 9:
```suggestion
      prefix: "chore"
```

The docs make it sound like the colon is added for us. No need to specify it.

levibostian (commented):
:excited:!

.github/workflows/dependabot.yml
levibostian, line 6:
```suggestion
    schedule:
      interval: "weekly"
    reviewers:
      - "customerio/mobile"
```

How about automatically adding mobile group as reviewers?

xtreem88: Yes, it is going to be weekly, I set it to hourly so that we can test it

levibostian (approved):

Issue #83, opened 2023-06-15
Push deep links not working when using multiple push notification modules (ie `expo-notifications`)


wanschi:
Hello,

I have integrated the Customer.io Expo SDK. It seems like everything is working. 

- I can start the permission request
- A token is issued
- I have turned on rich push
- I receive push messages (also with photos, so rich messages are enabled and configured).

What does not work though are deep links. Deep links "dry" launch the app too (tested with uri-scheme), but not with a Customer.io push notification. Also, Linking.useURL() is not triggered after I receive a push notification and launch the app. I always end up in the homescreen of the app. 

I have also tried debugging the whole thing via Swift. But it seems like it doesn't jump into the code for the deep links at all.

"expo": "~48.0.18",
"customerio-expo-plugin":"^1.0.0-beta.10",
"customerio-reactnative": "2.4.0",

(Unfortunately, the latest version of customerio-reactnative does not work for us).

Thank you

ami-aman:
Hi @wanschi 

Thanks for reaching out and apologies for the inconvenience caused. 

I would like to let you know that this is a known issue with our expo plugin. We are actively working on improving our plugin, but I'm unable to provide a specific timeline for the resolution at this moment. Rest assured I will make sure to raise this with our team internally and get the fix out of the door soon. Your patience and understanding are greatly appreciated.

Meanwhile, let us know if you have any other issues.

heuels:
Following! We at @on-deck would also love to use deep links from the CIO push notifications.

dgsunesen:
Same here. Deep link support would be awesome.

@wanschi if I may ask you a question? Do you receive Customer.io push notifications when your app is foregrounded? Currently I only receive them when the app is backgrounded.

wanschi:
@dgsunesen Yes, same here. We're only receive them in the background. After opening them you'll end up on the first screen. In our case this is the left item in the root tab bar. 

@ami-aman Can you give us a rough estimation. Since our plan is to release in 3 weeks we'll need to know if we need to switch to another push notification solution.

dgsunesen:
@ami-aman Just FYI, I'm in contact with your Technical Support Engineers who promised me to reach out to you guys (Mobile team) on an ETA on when this could be fixed :) 

We're also planning on releasing a new version of app including Customer.io, but not being able to receive push notifications while app is foregrounded as well as deep links not working properly could be quite the party killer for us :(

ami-aman:
[...]

PR #112, opened 2023-11-16
feat: automatically handle push click events and increase compatibility with 3rd party push modules

levibostian:
Part of: https://github.com/customerio/issues/issues/11289

# Goals of PR

- Install new feature in iOS SDK of automatic push click handling. This is done by installing a specific version of the RN SDK that has this iOS feature installed. 
- Remove all native iOS code modifications that previously handled push click handling. 
- Add compatibility with `expo-notifications` package. Closes: https://github.com/customerio/customerio-expo-plugin/issues/83

# Blocked by
- Deploy native iOS https://github.com/customerio/customerio-ios/pull/403
- Update RN SDK to include ^^^ iOS feature. 

# Docs 

[PR](https://github.com/customerio/docs/pull/1670)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": ">=3.X.Y"
+    "customerio-reactnative": ">=3.4.0"
   },
   "devDependencies": {
     "@expo/config-plugins": "^4.1.4",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": "^3.0.0"
+    "customerio-reactnative": ">=3.X.Y"
   },
   "devDependencies": {
     "@expo/config-plugins": "^4.1.4",
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -85,9 +85,7 @@ export const CIO_CONFIGURECIOSDKPUSHNOTIFICATION_SNIPPET = `
 
 export const CIO_INITIALIZECIOSDK_SNIPPET = `  
   [pnHandlerObj initializeCioSdk];
-`;
 
-export const CIO_CONFIGURECIOSDKUSERNOTIFICATIONCENTER_SNIPPET = `
 // Code to make the CIO SDK compatible with expo-notifications package.
 // 
 // The CIO SDK and expo-notifications both need to handle when a push gets clicked. However, iOS only allows one click handler to be set per app.
--- a/src/helpers/constants/ios.ts
+++ b/src/helpers/constants/ios.ts
@@ -9,7 +9,7 @@ pluginPackageRoot = path.dirname(pluginPackageRoot);
 export const LOCAL_PATH_TO_RN_SDK = path.join(
   pluginPackageRoot,
   '../customerio-reactnative'
-)
+);
 
 export const LOCAL_PATH_TO_CIO_NSE_FILES = path.join(
   pluginPackageRoot,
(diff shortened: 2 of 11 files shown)


src/helpers/utils/codeInjection.ts
```diff
@@ -61,5 +61,5 @@ export function injectCodeByLineNumber(
     content = [...lines.slice(0, index), snippet, ...lines.slice(index)];
   }
 
-  return content;
+  return content.join('\n');
```
levibostian, line 64: All other `injectN` functions in this file return a `string` while this `inject` function returned an `Array`. I struggled using this function during development until I made this change.

src/ios/withAppDelegateModifications.ts
```diff
@@ -62,7 +51,7 @@ const addImport = (stringContents: string, appName: string) => {
     stringContents,
     endOfMatchIndex,
     addedImport
-  ).join('\n');
+  );
```
levibostian, line 54: Removed this after `injectCodeByLineNumber` now returns a `string` instead of `Array`.

package.json
```diff
@@ -40,7 +40,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": "^3.0.0"
+    "customerio-reactnative": ">=3.X.Y"
```
levibostian, line 43: This needs to be updated after new version of RN SDK gets deployed. The RN SDK will need to install newest version of iOS SDK that includes [this PR](https://github.com/customerio/customerio-ios/pull/403).

levibostian:
I have a modification to make to this PR. Changing to draft until this change is done.

src/types/cio-types.ts
```diff
@@ -16,7 +16,11 @@ export type CustomerIOPluginOptionsIOS = {
   appleTeamId?: string;
   appName?: string;
   disableNotificationRegistration?: boolean;
+  /**
+   * @deprecated No longer has any effect. Use autoTrackPushEvents to control if push metrics should be automatically tracked by SDK.
+   */
   handleNotificationClick?: boolean;
```
levibostian, line 22: `handleNotificationClick` was an expo config plugin option because some customers would need to disable it if they were using another SDK that performs click handling (such as `expo-notifications`). That problem has been resolved by the new iOS feature and so there is no reason to disable the behavior.

src/types/cio-types.ts
```diff
@@ -16,7 +16,11 @@ export type CustomerIOPluginOptionsIOS = {
...
+  autoTrackPushEvents?: boolean;
```
levibostian, line 23: Added new expo plugin config option which maps to the native iOS `autoTrackPushEvents` config option.

src/helpers/native-files/ios/PushService.swift
```diff
@@ -11,6 +11,18 @@ public class CIOAppPushNotificationsHandler : NSObject {
 
   {{REGISTER_SNIPPET}}
 
+  @objc(initializeCioSdk)
+  public func initializeCioSdk() {
```
Shahroz16, line 15: I believe the expo utilized React native SDK initialization as well, can that be an issue?

Shahroz16 (commented):
added my thoughts and it looks okay, but would appreciate more eyes from folks with expo experience.

src/helpers/constants/ios.ts
```diff
@@ -42,20 +42,22 @@ export const CIO_APPDELEGATEDECLARATION_REGEX =
...
+  /^\s*RCTBridge\s*\*\s*\w+\s*=\s*\[\s*self\.reactDelegate\s+createBridgeWithDelegate:self\s+launchOptions:launchOptions\s*\];\s*$/gm;
 
-export const CIO_LAUNCHOPTIONS_DEEPLINK_MODIFIEDOPTIONS_REGEX = 
-/^\s*return\s\[\s*super\s*application:\s*application\s*didFinishLaunchingWithOptions\s*:\s*launchOptions\s*\];/gm;
+export const CIO_LAUNCHOPTIONS_DEEPLINK_MODIFIEDOPTIONS_REGEX =
+  /^\s*return\s\[\s*super\s*application:\s*application\s*didFinishLaunchingWithOptions\s*:\s*launchOptions\s*\];/gm;
 
-export const CIO_DEEPLINK_COMMENT_REGEX = /\sDeep link workaround for app killed state start/gm;
+export const CIO_DEEPLINK_COMMENT_REGEX =
+  /\sDeep link workaround for app killed state start/gm;
```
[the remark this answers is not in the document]
levibostian: Ah, looks like latest `beta` needs merged in. Done.

src/helpers/native-files/ios/PushService.swift
levibostian: Yes, we do suggest Expo customers to initialize the SDK via RN SDK. 

The only issue that I am aware of was a memory issue that was caused by native initialize() being called multiple times in a short amount of time. That issue did receive a [bug fix](https://github.com/customerio/customerio-reactnative/pull/217) recently that should prevent that from happening.

package.json
ami-aman: Isn't this syntax same as writing "customerio-reactnative": ">=3.0.0"? If you want to go with the above syntax then I think you should test it in an expo app before merging.

src/helpers/constants/ios.ts
```diff
@@ -83,9 +83,32 @@ export const CIO_CONFIGURECIOSDKPUSHNOTIFICATION_SNIPPET = `
...
+# if __has_include(<EXNotifications/EXNotificationCenterDelegate.h>)
+  // Creating a new instance, as the comments in expo-notifications suggests, does not work. With this code, if you send a CIO push to device and click on it,
+  // no push metrics reporting will occur.
+  // EXNotificationCenterDelegate *notificationCenterDelegate = [[EXNotificationCenterDelegate alloc] init];
+
+  // ...instead, get the singleton reference from Expo. 
+  id<UNUserNotificationCenterDelegate> notificationCenterDelegate = (id<UNUserNotificationCenterDelegate>) [EXModuleRegistryProvider getSingletonModuleForClass:[EXNotificationCenterDelegate class]];
   UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
-  center.delegate = self;
+  center.delegate = notificationCenterDelegate;
```
ami-aman, line 110: Has this been tested on an expo app ? I think I faced same issues even after setting this delegate. Let me know if you need help in testing this.

package.json
levibostian: Sorry, I don't understand what you're suggesting. 

This PR is changing the file to `>=3.X.Y` (will change `.X.Y` after iOS feature released) and from your comment, it seems that you agree that `>=` should be used. Do you have a suggestion for using something else in this PR?

src/helpers/constants/ios.ts
levibostian: Yes it has been tested. Both by our sample app and a customer testing it in their app.

package.json
ami-aman: I could have done a better job at explaining my previous comment. I understand that the version ranges `>=3.X.Y ` and `>=3.0.0` are equivalent, both indicating that any version of the package equal to or greater than `3.0.0` is acceptable and will be installed. Since you mentioned that you have not tested `>=3.X.Y` syntax, I wanted to verify with you if we are on the same page and the installation too gets tested before we release the fix.

I hope my explanation is more helpful this time.

ami-aman (approved):

PR #114, opened 2023-11-16
feat: show a push while app in foreground

levibostian:
Enables [this feature being added to iOS SDK](https://github.com/customerio/customerio-ios/pull/407) of configuring the SDK's behavior with automated push click handling. 

By default, pushes will show while app is in the foreground, just like on native iOS. 

Merges into: https://github.com/customerio/customerio-expo-plugin/pull/112

### DIFF
--- a/src/types/cio-types.ts
+++ b/src/types/cio-types.ts
@@ -17,6 +17,7 @@ export type CustomerIOPluginOptionsIOS = {
   appName?: string;
   disableNotificationRegistration?: boolean;
   handleNotificationClick?: boolean;
+  showPushAppInForeground?: boolean;
   handleDeeplinkInKilledState?: boolean;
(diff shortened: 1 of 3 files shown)


src/types/cio-types.ts
```diff
@@ -17,6 +17,7 @@ export type CustomerIOPluginOptionsIOS = {
   appName?: string;
   disableNotificationRegistration?: boolean;
   handleNotificationClick?: boolean;
+  showPushAppInForeground?: boolean;
```
levibostian, line 20: Added new expo plugin config option to match the iOS native SDK config option.

ami-aman (approved):
PR changes look good to me.

@levibostian Before merging if you could confirm that you have tested the changes in an Expo app?

PR #116, opened 2023-12-11
feat: disable the feature, if a customer needs to

levibostian:
adds ios feature https://github.com/customerio/customerio-ios/pull/430 to the expo plugin


levibostian:
Changing this to a draft PR. We may not need to modify Expo plugin at all depending on what config option we end up using in iOS PR https://github.com/customerio/customerio-ios/pull/430

levibostian:
Closing. We are re-using an existing expo config option to enable/disable feature. No change needed to expo plugin.

Issue #118, opened 2023-12-14
Rename NotificationService to CIONotificationService


darnfish:
I'm encountering issues using this Expo plugin in parallel with [OneSignal's Expo plugin](https://github.com/OneSignal/onesignal-expo-plugin), most notibly because both libraries use `NotificationService` as the name of the Objective C class. Do you know if it's possible to rename to CIONotificationService to avoid conflicts with OneSignal / any other config plugins?

ami-aman:
Hi @darnfish 

Thanks for reaching out and apologies for the inconvenience caused. To better assist you with displaying rich push notifications using both plugins, I have a few questions. Could you please tell me: When you run both plugins in your app, does `OneSignal` attempt to create a new target or add the files to the target that `customerio-expo-plugin` creates ?

darnfish:
Yeah exactly, that's the issue. I used `patch-package` to rename every instance of NotificationService to CIONotificationService and the build went through successfully. Here is the patch I created: https://gist.github.com/darnfish/a93abf72fc1bacd453dad778dd149e7c.

darnfish:
This was the CI error when this issue was occurring:
```
Generating Pods project
[!] An error occurred while processing the post-install hook of the Podfile.
[Xcodeproj] Consistency issue: no parent for object `NotificationService.m`: `SourcesBuildPhase`, `SourcesBuildPhase`
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project/object/helpers/groupable_helper.rb:19:in `parent'
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project/object/build_file.rb:77:in `ascii_plist_annotation'
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project.rb:297:in `block in to_ascii_plist'
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project.rb:296:in `each'
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project.rb:296:in `to_ascii_plist'
/Users/expo/.gems/arm64/gems/xcodeproj-1.23.0/lib/xcodeproj/project.rb:365:in `block in save'
/Users/expo/.gems/arm64/gems/atomos-0.1.3/lib/atomos.rb:21:in `block in atomic_write'
[...]

PR #119, opened 2023-12-14
fix: update nse class name to avoid conflicts with other libraries

ami-aman:
helps: https://github.com/customerio/customerio-expo-plugin/issues/118

The customer is having issues when multiple libraries have same class name such as `NotificationService` in NotificationServiceExtension target. The customer specifically mentioned about OneSignal and our expo plugin showing conflicts because of the same class name.


darnfish:
I think you might need to rename the actual file names too :~) but looking good - thank you so much!

ami-aman:
Reason of closing the PR - https://github.com/customerio/customerio-expo-plugin/issues/118#issuecomment-1866448559

Issue #131, opened 2024-03-18
Can't get IOS notifications push token


Samox:
Hello,
I have a infinite promise when I try to get the DeviceToken on iOS. Very similar to that issue [https://github.com/expo/expo/issues/16622](https://github.com/expo/expo/issues/16622).

I tried to add try/catch around the the `await nativeTokenPromise` but it never resolves nor fails.

I don't know what else to try nor how to debug this.
The problem disappears if I remove the plugin configuration.
I have tried with richpush: true and false. With disableNotificationRegistration true and false.

```ts
[
        'customerio-expo-plugin',
        {
          android: {
            googleServicesFile: './google-services.json',
          },
          ios: {
            pushNotification: {
              useRichPush: false,
              env: {
                siteId: process.env.EXPO_PUBLIC_CUSTOMERIO_SITE_ID,
                apiKey: process.env.EXPO_PUBLIC_CUSTOMERIO_API_KEY,
                region: 'eu',
              },
            },
          },
        },
      ],
```

I use

```json
"expo": "^50.0.6",  
"customerio-expo-plugin": "^1.0.0-beta.15",
"customerio-reactnative": "^3.5.2",
"expo-notifications": "~0.27.6",
```

levibostian:
Oh, sorry for the issue being auto-closed. I'm new to using Linear.

Samox:
Hello thanks for your quick response!
My answers for your questions:
- Yes the notifications are enabled 
![CleanShot 2024-03-19 at 10 10 10@2x](https://github.com/customerio/customerio-expo-plugin/assets/6906559/a3c2fc97-a826-497d-b2af-341f4e94af67)


- The flow is we first login the user, then prompt to accept the notifications. After login this code runs.

```ts
 if (Device.isDevice) {
    const { status: existingStatus } = await Notifications.getPermissionsAsync();
    let finalStatus = existingStatus;
    if (existingStatus !== 'granted') {
      const { status } = await Notifications.requestPermissionsAsync();
      finalStatus = status;
    }
    if (finalStatus !== 'granted') {
      console.warn('Failed to get push token for push notification!',  finalStatus);
    }
    console.log('This works', finalStatus);

    const expoPushToken = await Notifications.getExpoPushTokenAsync({
      projectId: Constants.expoConfig.extra.eas.projectId,
    }); -> We are stuck here. Promise is not resolved, and does not throw either
    

    console.log(`Push token: ${expoPushToken.data}`);
```

[...]

PR #132, opened 2024-03-20
feat: add autoFetchDeviceToken option for ios

Samox:


levibostian:
Thanks for opening a PR, Samox! Appreciate your time and hard work to contribute some code to the project. 

As mentioned [in this recent comment](https://github.com/customerio/customerio-expo-plugin/issues/131#issuecomment-2010139270), Expo apps should not need to change the value of `autoFetchDeviceToken` unless there is a problem. 

With that in mind, I'll close this PR for now and will instead provide a bug fix to the problem that you're experiencing. 

Thanks!

PR #161, opened 2024-11-25
chore: Update plugin infrastructure to allow local/automation testing

mahmoud-elmorabea:
This PR serves as the first step to execute on this [tech proposal](https://www.notion.so/custio/Tech-Proposal-Reliable-Expo-plugin-1454302f4c2b800fb7d3c30cd76ffbfd)

The PR restructures our plugin so we can test it locally and write some automation tests to verify the native code that's being generated.

This PR only does the following:
- Moves the plugin into its own top level directory `plugin`
- Creates an empty `test-app` that we can use to test the plugin
- Creates the infrastructure and an example of a unit test

Following PRs will add more tests and hook test running to CI.


mahmoud-elmorabea:
Closed in favor of broken up PR #165, #166, #167

PR #164, opened 2024-11-26
chore: Add Gituhb action job to run plugin tests

mahmoud-elmorabea:
This PR adds a new job that runs unit tests for the Expo plugin when changes are pushed to branches and on PRs.

The new job will run all unit tests under the `__tests__` directory in this repository.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,3 +42,4 @@ jobs:
 
       - name: Run plugin tests
         run: npm run test-plugin
+        
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,3 +23,22 @@ jobs:
 
       - name: Publish test
         run: npm publish --dry-run
+
+  test-plugin:
+    name: Test unit tests for plguin
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -23,3 +23,22 @@ jobs:
...
+    name: Test unit tests for plguin
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install plugin dependencies
+        working-directory: plugin
```
Shahroz16, line 37: i don't think we need to specify this for all steps. if you need to run both jobs in the same directory, you can just do this 

```
defaults:
  run:
    working-directory: ./plugin
```

and if there is any step which probably needs a different directory, I don't think there should be, there you can specify `working-directory` if needed.

Shahroz16 (approved):

.github/workflows/test.yml
mahmoud-elmorabea: This is the only step that needs this `working-directory` and the rest of the steps happen at the root directory of the repository, that's why I only added it for this step.

PR #166, opened 2024-11-26
chore: Create an empty test app to be used to test the plugin locally

mahmoud-elmorabea:
This PR serves as the first step to execute on this [tech proposal](https://www.notion.so/custio/Tech-Proposal-Reliable-Expo-plugin-1454302f4c2b800fb7d3c30cd76ffbfd)

The PR restructures our plugin so we can test it locally and write some automation tests to verify the native code that's being generated.

This PR only does the following:
- Creates an empty `test-app` that we can use to test the plugin

The app is generated with `npx create-expo-app@latest test-app --template blank`


test-app/app.json
```diff
@@ -0,0 +1,50 @@
+{
+  "expo": {
+    "name": "test-app",
```
mrehan27, line 3: Can we use more descriptive name instead? Something like ExpoTestApp, Expo, CIOExpo, etc. so it's clearer and easier to identify, especially when distributing through Firebase in future?

test-app/app.json
```diff
@@ -0,0 +1,50 @@
...
+        {
+          "android": {
+            "googleServicesFile": "./files/google-services.json"
+          },
+          "ios": {
+            "pushNotification": {
+              "useRichPush": true,
+              "env": {
+                "siteId": "<siteId>",
+                "apiKey": "<apiKey>",
```
mrehan27, line 41: How are these values updated while testing locally or on CI?

test-app/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "ios": "expo run:ios",
+    "web": "expo start --web"
+  },
+  "dependencies": {
+    "expo": "~52.0.11",
+    "expo-status-bar": "~2.0.0",
+    "react": "18.3.1",
+    "react-native": "0.76.3",
+    "customerio-expo-plugin": "file:../plugin",
+    "customerio-reactnative": "4.0.1"
```
mrehan27, line 17: Any specific reason for not using the latest version? Also, since this is only for internal test app, we could also specify version range so it fetches the latest version automatically. That said, it's totally fine to stick with a specific version if there's a reason for it, but just wanted to check.

test-app/app.json
mahmoud-elmorabea: I think we can inject them using Github secrets for CI and maybe environment variables for local development

test-app/package.json
mahmoud-elmorabea: Nope, no reason, probably just a typo on my end, will use the latest one!

test-app/app.json
mahmoud-elmorabea: When we are at the point when we start adding stuff to the test-app and publish it, we can name it whatever we want 😃

mrehan27: I usually find it easier to set up basic info like app name, etc., when adding templates, especially when we know it's definitely being included and plan to improve it over time. It also helps ensure it's not overlooked later. That said, if you prefer to keep it for later, I'm fine with that too. But would appreciate if you can note it down somewhere in relevant docs or include in relevant ticket to make sure it doesn't get missed when the time comes.

test-app/app.json
mrehan27: I don't see any script or other mechanism to inject these values currently. And this might make local testing a bit challenging since we'd need to manually update the values each time and remember to revert them before committing. Or do you plan to address this later?

mahmoud-elmorabea: Yes, this will addressed later as part of testing and distributing the test-app

test-app/app.json
mahmoud-elmorabea: Yes, I think it's better to do this when setting up app distribution, captured in [this task](https://linear.app/customerio/issue/MBL-684/create-a-mechanism-to-internally-test-expo-plugin)

PR #169, opened 2024-11-29
chore: Create plugin iOS native code generation snapshot tests

mahmoud-elmorabea:
This PR adds unit tests for the native code generated by our Expo plugin.

This PR focuses on the iOS generated code.

##### Option 1: Use Jest's snapshot testing (Recommended)
- This option uses Jest's snapshot testing to capture a snapshot of the result of the generated code by our Expo plugin
- Then for all subsequent runs we use the captured snapshot to make sure that the generated code is still as expected

##### Option 2: Custom testing
- This option would depend on us writing tests that find certain code parts or lines that are generated
- This option is a bit problematic since we have to handle different versions of Expo by ourselves
- This option is also challenging since we'll probably have to write code that is equality complicated to the code that generate the code under test


Shahroz16 (commented):
Same as other PR, we will be discussing the changes sync. Adding comment because didn't want to leave PRs unattended.

__tests__/ios/app-delegate-header.test.js
Shahroz16, line 1: is there a specific reason the file name is lower letter here compared to others?

__tests__/ios/app-delegate-header.test.js
```diff
@@ -0,0 +1,12 @@
+const fs = require("fs-extra");
+const path = require("path");
+
+const testProjectPath = path.join(__dirname, "../../test-app");
+const iosPath = path.join(testProjectPath, "ios");
+const appDelegateHeaderPath = path.join(iosPath, "testapp/AppDelegate.h");
+
+test("Plugin injects CIO imports and calls into AppDelegate.h (Expo v52)", async () => {
```
Shahroz16, line 8: do we need `(Expo v52)` here? we are reading from 1 static path, `../../test-app` 

so we will be running that app, with different expo versions matrix I am assuming. So does adding  v52 here adds value?

__tests__/ios/app-delegate-impl.test.js
```diff
@@ -0,0 +1,12 @@
+const fs = require("fs-extra");
+const path = require("path");
+
+const testProjectPath = path.join(__dirname, "../../test-app");
+const iosPath = path.join(testProjectPath, "ios");
+const appDelegateImplPath = path.join(iosPath, "testapp/AppDelegate.mm");
+
+test("Plugin injects CIO imports and calls into AppDelegate.mm (Expo v52)", async () => {
```
Shahroz16, line 8: same as above one.

Shahroz16 (approved):

__tests__/ios/app-delegate-header.test.js
mahmoud-elmorabea: Not at all, just my own inconsistencies 😂 

I will rename them to `AppDelegate-header` and `AppDelegate-impl` to be consistent with other test files.

__tests__/ios/app-delegate-header.test.js
mahmoud-elmorabea: You are absolutely right! 

For files that are dependent on external frameworks, I am used to including the latest version that was used to generate the snapshots. So that if the test fails with version 53 for example, it kinda gives you an idea that the reason could be the new Expo version.

But they serve no functional purpose. I will remove the version to avoid confusion.

__tests__/ios/app-delegate-impl.test.js
mahmoud-elmorabea: [Same as this](https://github.com/customerio/customerio-expo-plugin/pull/169#discussion_r1879809037)

PR #170, opened 2024-11-29
chore: Create plugin Android native code generation snapshot tests

mahmoud-elmorabea:
This PR adds unit tests for the native code generated by our Expo plugin.

This PR focuses on the Android generated code.

##### Option 1: Use Jest's snapshot testing (Recommended)
- This option uses Jest's snapshot testing to capture a snapshot of the result of the generated code by our Expo plugin
- Then for all subsequent runs we use the captured snapshot to make sure that the generated code is still as expected

##### Option 2: Custom testing
- This option would depend on us writing tests that find certain code parts or lines that are generated
- This option is a bit problematic since we have to handle different versions of Expo by ourselves
- This option is also challenging since we'll probably have to write code that is equality complicated to the code that generate the code under test


Shahroz16 (commented):
Great work here, we will be discussing the changes sync. Adding comment because didn't want to leave PRs unattended.

__tests__/android/app-gradle-build.test.js
```diff
@@ -0,0 +1,12 @@
+const fs = require("fs-extra");
+const path = require("path");
+
+const testProjectPath = path.join(__dirname, "../../test-app");
+const androidPath = path.join(testProjectPath, "android");
+const appBuildGradlePath = path.join(androidPath, "app/build.gradle");
+
+test("Plugin applies Google Services plugin in the app Gradle build file (Expo v52)", async () => {
```
Shahroz16, line 8: the same comment as other PR, versioning of here would probably only make sense when we use approaches like of fixtures, like [here](https://github.com/invertase/react-native-firebase/blob/3623a504a6d07fa148390bd3f97d6c06fea8955d/packages/app/plugin/__tests__/iosPlugin.test.ts)?

Shahroz16 (approved):

__tests__/android/app-gradle-build.test.js
mahmoud-elmorabea: Absolutely right! Removing them to avoid confusion.

PR #172, opened 2024-12-10
chore: Build basic test app basic functionality

mahmoud-elmorabea:
This PR starts to build basic functionality for the Expo test app so we can verify E2E integration is working.

This is no way a complete or final version of the app, this is only to unblock testing and basic verification of Expo SDK functionality.

The test app with its current state should allow you to:
- Identify a user
- Clear user identification
- Send an event
- Request push permission to allow you to test push notifications

Screenshot 2024-12-10 at 3 48 29 PM

I've tested the following scenarios with the test app and the [Expo workspace](https://fly.customer.io/workspaces/94996/journeys/dashboard) and they working:
- ✅ Anonymous user tracking 
- ✅ Identify a user
- ✅  Clear user identification
- ✅  Send push messages (Push setup has to be done locally)
- ✅  Send in-app message

There are 3 main things that are still missing and will be added in following PRs:
- Push setup (download certificates)
- Inject API key and Site ID through CI secrets
- Add more functionality to configure and test more SDK features

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,4 +42,52 @@ jobs:
 
       - name: Run plugin tests
         run: npm run test-plugin
-        
+  
+  build-android-app:
+    name: Build and compile Android app to ensure correctness
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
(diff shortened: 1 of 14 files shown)


scotttwittrockcio:
One small note. 3/4 of the other repos have any testing apps in a top level folder called "Apps". Either now or when we upgrade this to a full test app we should update the folder name to be consistent across SDKs.

PR #178, opened 2024-12-16
chore: Create a CI job that builds and publishes the Android test app

mahmoud-elmorabea:
This PR creates a CI job that builds and publishes the Expo Android app to Firebase.


.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Setup CIO workspace credentials in test apps config
+        run: |
+          sd '@CDP_API_KEY@' "${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_CDP_API_KEY }}" test-app/app.json
+          sd '@SITE_ID@' "${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_SITE_ID }}" test-app/app.json
+
+      - name: Build Android app with fastlane
+        uses: maierj/fastlane-action@v3.1.0
+        with:
+          lane: android build
```
mrehan27, line 51: It looks like generated build will always use version `1.0.0 (1)`, which means we won't be able to update the app but will need to uninstall/reinstall it instead? Something we can improve later, but just wanted to confirm so we're aligned on the expectations here.

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,55 @@
...
+          subdirectory: test-app
+        env:
+          FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
+        continue-on-error: true
```
mrehan27, line 55: Why do we need `continue-on-error: true` here? If not needed, it is better to remove to ensure CI fails for any unexpected issues.

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,55 @@
+name: Publish test apps
+
+on: [push, pull_request]
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Fixed to be `beta` and branches under `feature` directory

.github/workflows/build-sample-apps.yml
mahmoud-elmorabea: Good point, fixed to align with other sample apps

.github/workflows/build-sample-apps.yml
mahmoud-elmorabea: I will keep this for now because I'm planning to add the iOS to the same job

mahmoud-elmorabea: Done

mrehan27 (approved):

PR #179, opened 2024-12-16
chore: Update plugin config to work with standalone apps

mahmoud-elmorabea:
This PR makes sure that our plugin works with both:
- Local development through using the plugin using relative path
- Standalone apps using the plugin through Github branches or published versions

The idea is that we maintain the plugin config at the root level and adjust our configurations and scripts to still make sure that the source code of the plugin and the generated build files live in the `plugin` sub-directory for better organization.

You can test the standalone setup by declaring the plugin dependency `package.json` in a standalone app as such:
`"customerio-expo-plugin": "github:customerio/customerio-expo-plugin#MBL-684-plugin-config-standalone-apps"`

I've also confirmed that when installing the plugin this way, only the required files are included in the `node_modules` directory
Screenshot 2024-12-17 at 10 36 43 AM


package.json
Shahroz16, line 1: this is just moving the packages right? nothing to see here?

package.json
```diff
@@ -1,17 +1,123 @@
...
     "cleanup": "bash scripts/cleanup-after-tests.sh",
     "test-plugin": "npm run setup-tests && npm run test && npm run cleanup"
   },
+  "keywords": [
+    "react-native",
+    "ios",
+    "android"
+  ],
+  "repository": "https://github.com/customerio/customerio-expo-plugin",
+  "author": "ami-segun <segun@customer.io> (https://customer.io/)",
```
Shahroz16, line 40: probably change it to mobile as we did for react native SDK?

Shahroz16 (approved):
Looks good, please verify the deployment scripts are also updated for releases. `node-modules` is already checked as per PR desrcription, I am assuming test app isn't getting included as part of npm release as well?

package.json
mahmoud-elmorabea: Yup, just merging it with the top level package.json, nothing changes

package.json
mahmoud-elmorabea: What do you mean by "mobile"? I checked the RN SDK and it says Aman in there, I'll go ahead and merge this PR and I'll take a note to update this before release

mahmoud-elmorabea:
Thanks for the review! @Shahroz16 

> Looks good, please verify the deployment scripts are also updated for releases

What do you mean by it's updated for releases?  If you mean the deployment scripts, can we test those without triggering an actual release?

I'll take a note of this as well, but I'll merge this for other PRs to be updated with it

PR #181, opened 2024-12-17
chore: Build and publish iOS test app to Firebase

mahmoud-elmorabea:
This PR adds a job to build and publish the Expo iOS test app to Firebase.

_I am pretty sure some parts of the flow are not optimal, but this should be good enough to unblock testing and we can optimize the scripts later on._

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -44,7 +44,7 @@ jobs:
         working-directory: plugin
         run: npm ci
 
-      - name: Setup test projects
+      - name: Setup test project
         run: npm run setup-tests
 
       - name: Setup CIO workspace credentials in test apps config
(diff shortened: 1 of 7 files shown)


scripts/add-google-service-ios.rb
```diff
@@ -0,0 +1,27 @@
+require 'xcodeproj'
```
mahmoud-elmorabea, line 1: This script makes sure that the `GoogleService-Info.plist` is added to the bundle correctly and to the xcode project file.

test-app/fastlane/Fastfile
```diff
@@ -55,3 +55,91 @@ platform :android do
...
+    )
+
+    firebase_app_distribution(
+      service_credentials_file: service_credentials_file_path,
+      groups: get_build_test_groups(),
+      release_notes: get_build_notes()
+    )
+  end
+
+  lane :download_ci_code_signing_files_local do |values|
```
mahmoud-elmorabea, line 130: I made a copy of this file because the original lane downloads two profiles (Development, Distribution). Gym kept picking the Development version and failing no matter what I tried 😄 

I choose to make this copy to only download Distribution to overcome that issue. 

_I don't think this is the best solution which we can probably make better in the future_

.github/workflows/build-sample-apps.yml
```diff
@@ -60,3 +60,59 @@ jobs:
...
+      - name: Bundle install
+        working-directory: test-app
+        run: bundle install
+
+      - name: Install plugin dependencies
+        working-directory: plugin
+        run: npm ci
+
+      - name: Setup test project
+        run: npm run setup-tests
```
Shahroz16, line 98: assuming this will run test, its okay if we don't update it now but maybe lets add a todo to have its own separate workflow or job for tests?

test-app/fastlane/Fastfile
```diff
@@ -55,3 +55,91 @@ platform :android do
...
+      export_method: "ad-hoc",
+      configuration: "Release",
+      codesigning_identity: "Apple Distribution: Peaberry Software, Inc. (2YC97BQN3N)",
+      export_options: {
+        provisioningProfiles: {
+          "io.customer.ami" => "match AdHoc io.customer.ami",
+          "io.customer.ami.richpush" => "match AdHoc io.customer.ami.richpush"
+        },
+        signingStyle: "manual",
+        teamID: "2YC97BQN3N"
```
Shahroz16, line 110: we should ideally inject these

Shahroz16 (approved):
Happy to go forward with this, but we should definitely create a follow up linear ticket so we don't forget about the optimizations and approach.

.github/workflows/build-sample-apps.yml
mahmoud-elmorabea: I'm just reusing the script that "setups" up tests, it just builds the plugin and native apps, it doesn't actually run the tests. I will add a follow up to rename it

PR #182, opened 2024-12-18
chore: fix user agent

mrehan27:
part of [MBL-541](https://linear.app/customerio/issue/MBL-541/expo-user-agent)

### Changes

- Updated the plugin to include SDK info in meta-data for Android apps
- Added helper method to extract plugin version information
- Updated post-install script to fix React Native package path for iOS, which was affected by moving files into `plugin` directory
- Minor code cleanup in related files


mrehan27:
Closing this in favor of #184 as updating `AndroidManifest` approach is failing due to issues with merging manifests. Added more details to related Linear ticket.

PR #194, opened 2024-12-24
chore: Create workflow to unpublish unwanted plugin npm versions

mahmoud-elmorabea:
This PR introduces a workflow to unpublish plugin versions that we might have mistakenly published.


Shahroz16 (approved):
Looks good as long as NPM lets you delete the packages or on publish it.

PR #211, opened 2025-02-08
chore: Public release builds

mahmoud-elmorabea:
This PR creates a new workflow that publishes a test app that uses the latest published `customerio-expo-plugin` version. This workflow can be launched manually.

The published test apps on Android and iOS have the same version as the SDK version they were built with to make it easy to find the desired version for the tester.

There can be multiple test app versions on the Firebase Distribution `public` channel with the same version in case we trigger this workflow for the same SDK version but with new changes to the test apps.


__tests__/android/__snapshots__/main-gradle-build.test.js.snap
```diff
@@ -9,7 +9,7 @@ buildscript {
         minSdkVersion = Integer.parseInt(findProperty('android.minSdkVersion') ?: '24')
         compileSdkVersion = Integer.parseInt(findProperty('android.compileSdkVersion') ?: '35')
         targetSdkVersion = Integer.parseInt(findProperty('android.targetSdkVersion') ?: '34')
-        kotlinVersion = findProperty('android.kotlinVersion') ?: '1.9.24'
+        kotlinVersion = findProperty('android.kotlinVersion') ?: '1.9.25'
```
mahmoud-elmorabea, line 12: Updated from npm dependencies update

scripts/applyLocalEnvValues.js
```diff
@@ -0,0 +1,60 @@
+const fs = require('fs');
+const dotenv = require('dotenv');
+const { execSync } = require('child_process');
+
+function installPluginTarball() {
```
mahmoud-elmorabea, line 5: We can use this script to update the project with values from a `local.env`, at the moment this only reads `sdkVersion` but we can add support for API key and site ID like other SDKs

test-app/.gitignore
```diff
@@ -6,6 +6,7 @@ android/
 
 # dependencies
 node_modules/
+package-lock.json
```
mahmoud-elmorabea, line 9: Since the sample app now can be built with either the local `tgz` or a package version from NPM, we cannot check in this lock file because it will have either one of them specified.

test-app/fastlane/Fastfile
```diff
@@ -130,7 +130,7 @@ lane :update_expo_test_app_version do |values|
     UI.user_error!("app.json not found at #{app_json_path}")
   end
 
-  version_name = get_new_app_version()
+  version_name = values[:sdk_version] || get_new_app_version()
```
mahmoud-elmorabea, line 133: Use SDK version as app version otherwise, use the logic for all other sample apps

Shahroz16 (approved):
lgtm

test-app/.gitignore
mrehan27: I think we can still keep the lock file. For both tgz and exact dependencies, `package-lock` can be temporarily updated on CI and should work fine. If possible, I'd prefer not adding it to `.gitignore` as it could create gaps between local dev environments and lead to more "works on my machine" issues 😅

mrehan27 (approved):

test-app/.gitignore
mahmoud-elmorabea: As far as I've seen in my testing, running `npm install` should update the `package-lock` file but that doesn't always work, sometimes you have to manually delete it and re-run the command for it to work.

I think it should be fine to ignore it because we do want dependencies to keep updated (within the limits we have defined, minor and patch updates) when running on CI so we catch broken functionality with minor dependencies updates.

I was intentionally using `npm install` [here](https://github.com/customerio/customerio-expo-plugin/blob/beta/scripts/setup-test.sh#L11) for that reason, instead of `npm ci`

PR #217, opened 2025-03-03
chore: FCM base implementation

mahmoud-elmorabea:
This PR adds the basic support for customers to be able to use FCM as push provider.


plugin/src/helpers/utils/injectCIOPodfileCode.ts
```diff
@@ -51,11 +53,12 @@ export async function injectCIONotificationPodfileCode(
   const matches = podfile.match(new RegExp(blockStart));
 
   if (!matches) {
+    const useFcm = pushProvider === 'fcm';
     const snippetToInjectInPodfile = `
 ${blockStart}
 target 'NotificationService' do
   ${useFrameworks === 'static' ? 'use_frameworks! :linkage => :static' : ''}
-  pod 'customerio-reactnative-richpush/apn', :path => '${getRelativePathToRNSDK(
+  pod 'customerio-reactnative-richpush/${useFcm ? "fcm" : "apn"}', :path => '${getRelativePathToRNSDK(
```
mahmoud-elmorabea, line 61: I plan to create a global util to centralize the logic to check if push provider is FCM or APN

plugin/src/ios/withAppDelegateModifications.ts
```diff
@@ -138,6 +138,16 @@ const addExpoNotificationsHeaderModification = (stringContents: string) => {
   return stringContents;
 };
 
+const addFirebaseDelegateForwardDeclarationIfNeeded = (stringContents: string) => {
+  stringContents = injectCodeByLineNumber(
+    stringContents,
+    0,
+    '@protocol FIRMessagingDelegate;'
```
mahmoud-elmorabea, line 145: This solves an issue we had caused by `FIRMessagingDelegate` not being found during compilation.

We are adding this "Forward Declaration" here to indicate to the compiler that this type will be found later during linking/runtime.

Thanks @Ahmed-CIO for the suggestion!

test-app/app.json
```diff
@@ -77,6 +80,14 @@
             }
           }
         }
+      ],
+      [
+        "expo-build-properties",
+        {
+          "ios": {
+            "useFrameworks": "static"
```
mahmoud-elmorabea, line 88: Recommended [setup from Firebase](https://firebase.google.com/docs/ios/link-firebase-static-dynamic).

plugin/src/helpers/native-files/ios/fcm/Env.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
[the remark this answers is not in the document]
Shahroz16: hmm i do understand where you are coming from but this doesn't seem sustainable. like we are not just duplicating variable files (which changes on basis of platform), we are also duplicating static files.

plugin/src/types/cio-types.ts
```diff
@@ -25,6 +25,8 @@ export type CustomerIOPluginOptionsIOS = {
   handleDeeplinkInKilledState?: boolean;
   useFrameworks?: 'static' | 'dynamic';
   pushNotification?: {
+    provider?: string;
```
Shahroz16, line 28:
```suggestion
    provider?: 'apn' | 'fcm';
```

Shahroz16 (commented):
implementation looks good, i think we can avoid the duplication. We can definitely remove the duplication of static files, for files with variable code, I think we can have a better rendering/templating system. It would still involve regex but we can improve it. 


something like 

```
// 1. Define provider configurations
const providerConfigs = {
  apn: {
    import: 'CioMessagingPushAPN',
    initCode: 'MessagingPushAPN.initialize(...)'
  },
  fcm: {
    import: 'CioMessagingPushFCM',
    initCode: 'FirebaseApp.configure(); MessagingPushFCM.initialize(...)'
  }
};

// 2. Create a template file (NotificationService.template)
// import {{PROVIDER_IMPORT}}
// 
// function initialize() {
//   {{INIT_CODE}}
// }

// 3. Implement template rendering
function renderTemplate(template, provider) {
  const config = providerConfigs[provider];
  return template
    .replace('{{PROVIDER_IMPORT}}', config.import)
    .replace('{{INIT_CODE}}', config.initCode);
}

// 4. Generate provider-specific files
function generateFile(templatePath, provider, outputPath) {
  const template = readFile(templatePath);
  const rendered = renderTemplate(template, provider);
  writeFile(outputPath, rendered);
}
```

plugin/src/helpers/native-files/ios/fcm/Env.swift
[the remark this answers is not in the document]
Shahroz16: think about this way, i am adding config for `automaticDeviceToken` and now I need to add same line on both files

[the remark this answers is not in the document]
Shahroz16: Okay, so lets say, we stick with duplication to avoid templating and rendering, but why do we need to duplicate those files which are just static and doesn't need any regex for them? for example `Env.swift` , `NotificationService.h` etc?

We can have a common folder and have these files in there? and duplicated files depending on the platform can still exist for files where code is dependent on platform?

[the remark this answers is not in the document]
Shahroz16: Okay, so lets say, we stick with duplication to avoid templating and rendering, but why do we need to duplicate those files which are just static and doesn't need any regex for them? for example Env.swift , NotificationService.h etc?

We can have a common folder and have these files in there? and duplicated files depending on the platform can still exist for files where code is dependent on platform?

Shahroz16 (commented):
in order to remove back and forth and speed up things, I'll create a PR with duplication removed of alteast static files, like I am suggesting. 

For this PR, can we look at why the tests are failing and also what are we planning to do with temporary pointing test app to FCM?

plugin/src/ios/utils.ts
```diff
@@ -0,0 +1,10 @@
+import type { CustomerIOPluginOptionsIOS } from '../types/cio-types';
+
+/**
+ * Returns t
+ * @param iosOptions The plugin iOS configuration options
+ * @returns true if FCM is configured to be used as push provider
+ */
+export const isFcmPushProvider = (iosOptions?: CustomerIOPluginOptionsIOS): boolean => {
+    return iosOptions?.pushNotification?.provider === "fcm";
```
mrehan27, line 9: We should make this case insensitive to prevent issues from customer typos, like using `FCM` instead of `fcm`.

mrehan27 (approved):
Changes look good, considering the follow ups we discussed on call.

PR #221, opened 2025-03-06
chore: Build iOS sample app for FCM and APN

mahmoud-elmorabea:
This PR is a quality of life improvement for mobile developers and internal testers by publishing two iOS versions from CI one for APN and FCM, so developers don't have to manually change the config to test both variants.

This PR builds on the approach we used to inject SDK version through `local.env`. An additional property is added to `local.env` for `pushProvider`. This approach could work locally as well, if you create `local.env` locally and populate it with values it will make the correct configuration for you.

### DIFF
--- a/.github/workflows/reusable_build_sample_apps.yml
+++ b/.github/workflows/reusable_build_sample_apps.yml
@@ -139,6 +139,9 @@ jobs:
   build-ios-app:
     name: Build and publish iOS app
     runs-on: macos-14
+    strategy:
+      matrix:
+        push_provider: [fcm, apn]
     env:
       GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64: ${{ secrets.GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64 }}
       FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
@@ -185,10 +188,13 @@ jobs:
           echo "firebase_distribution_groups=$(IFS=','; echo "${distribution_groups[*]}")" >> $GITHUB_ENV
 
       - name: Setup local.env file for sample app
-        if: ${{ inputs.use_latest_sdk_version == true }}
+        shell: bash
         run: |
           touch "test-app/local.env"
-          echo "sdkVersion=${{ env.LATEST_TAG }}" >> "test-app/local.env"
+          if [ "${{ inputs.use_latest_sdk_version }}" == "true" ]; then
+            echo "sdkVersion=${{ env.LATEST_TAG }}" >> "test-app/local.env"
+          fi
+          echo "pushProvider=${{ matrix.push_provider }}" >> "test-app/local.env"
 
       - name: Install XCode
         uses: maxim-lobanov/setup-xcode@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -229,15 +235,21 @@ jobs:
...
-          sd "\"siteId\".*" "\"siteId\": \"${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_SITE_ID }}\"," "$APP_JSON_FILE"
           sd "\"workspaceName\".*" "\"workspaceName\": \"Mobile: Expo\"," "$APP_JSON_FILE"
           sd "\"branchName\".*" "\"branchName\": \"${{ env.BRANCH_NAME }}\"," "$APP_JSON_FILE"
           sd "\"commitHash\".*" "\"commitHash\": \"${{ env.COMMIT_HASH }}\"," "$APP_JSON_FILE"
           LAST_TAG="${LATEST_TAG:-untagged}"
           COMMITS_AHEAD=$(git rev-list $LAST_TAG..HEAD --count 2>/dev/null || echo "untracked")
           sd "\"commitsAheadCount\".*" "\"commitsAheadCount\": \"$COMMITS_AHEAD\"," "$APP_JSON_FILE"
 
+          if [ "${{ matrix.push_provider }}" == "fcm" ]; then
+            sd "\"cdpApiKey\".*" "\"cdpApiKey\": \"${{ secrets.TMP_CUSTOMERIO_EXPO_FCM_WORKSPACE_CDP_API_KEY }}\"," "$APP_JSON_FILE"
```
mrehan27, line 246: I think we can now rename these secrets to proper names and remove `TMP` from them.

.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -229,15 +235,21 @@ jobs:
...
+            sd "\"siteId\".*" "\"siteId\": \"${{ secrets.TMP_CUSTOMERIO_EXPO_FCM_WORKSPACE_SITE_ID }}\"," "$APP_JSON_FILE"
+          else
+            sd "\"cdpApiKey\".*" "\"cdpApiKey\": \"${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_CDP_API_KEY }}\"," "$APP_JSON_FILE"
```
mrehan27, line 249: It would be nice to rename these secrets to include `APN` or `DEFAULT` for better clarity.

scripts/applyLocalEnvValues.js
```diff
@@ -17,44 +17,97 @@ function installPluginTarball() {
...
+function updatePushProvider() {
+  const pushProvider = process.env.pushProvider;
+
+  if (!pushProvider) {
+    return;
+  }
 
-const packageJsonPath = `${testAppPath}/package.json`;
-const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+  const testAppPath = '../test-app';
```
mrehan27, line 61: We can probably remove this and use `testAppPath` defined at file level instead?

scripts/applyLocalEnvValues.js
```diff
@@ -17,44 +17,97 @@ function installPluginTarball() {
...
+
+  if (customerioPlugin) {
+    const pluginConfig = customerioPlugin[1];
+
+    if (pluginConfig.ios && pluginConfig.ios.pushNotification) {
+      if (pushProvider === 'fcm') {
+        // Update the provider value to "fcm"
+        pluginConfig.ios.pushNotification.provider = 'fcm';
+        pluginConfig.ios.pushNotification.googleServicesFile =
+          './files/google-services.json';
```
mrehan27, line 79: Doesn't iOS require `GoogleService-Info.plist` instead?

test-app/fastlane/Fastfile
```diff
@@ -102,7 +102,7 @@ platform :ios do
     firebase_app_distribution(
       service_credentials_file: service_credentials_file_path,
       groups: get_build_test_groups(distribution_groups: arguments[:distribution_groups]),
-      release_notes: get_build_notes()
+      release_notes: get_build_notes() + "\nPush provider: #{arguments[:push_provider]}"
```
mrehan27, line 105: Would this add `null` for public builds? We can either pass a value for public builds or set `APN` as default instead.

.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -229,15 +235,21 @@ jobs:
         run: |
           APP_JSON_FILE="test-app/app.json"
           sd "\"buildTimestamp\": .*" "\"buildTimestamp\": $(date +%s)," "$APP_JSON_FILE"
-          sd "\"cdpApiKey\".*" "\"cdpApiKey\": \"${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_CDP_API_KEY }}\"," "$APP_JSON_FILE"
-          sd "\"siteId\".*" "\"siteId\": \"${{ secrets.CUSTOMERIO_EXPO_WORKSPACE_SITE_ID }}\"," "$APP_JSON_FILE"
           sd "\"workspaceName\".*" "\"workspaceName\": \"Mobile: Expo\"," "$APP_JSON_FILE"
```
mrehan27, line 238: Can we update workspace name as well to make it easier to find the workspace?

scripts/applyLocalEnvValues.js
```diff
@@ -17,44 +17,97 @@ function installPluginTarball() {
...
+    }
+  } else {
+    console.error(
+      "'customerio-expo-plugin' not found in app.json, cannot update push provider config!"
     );
   }
-} else {
+
+  // Save the updated app.json
+  fs.writeFileSync(appJsonPath, JSON.stringify(appJson, null, 2) + '\n');
```
mrehan27, line 95: Also, we haven't updated `bundleIdentifier`. Isn't that needed for FCM app? Can the same Firebase app work for both FCM and APN?

scripts/applyLocalEnvValues.js
mahmoud-elmorabea: Good catch!

test-app/fastlane/Fastfile
mahmoud-elmorabea: No it wouldn't, all iOS apps will be built twice for APN and FCM even the public builds

scripts/applyLocalEnvValues.js
mahmoud-elmorabea: Yea sure, we can use the same app for both. FCM basically just uses the same APN setup under the hood

.github/workflows/reusable_build_sample_apps.yml
mahmoud-elmorabea: Not sure I understand?

scripts/applyLocalEnvValues.js
mrehan27: Yeah, I get that. I was just curious how same workspace can work with both APN and FCM. I tested it and noticed we're reusing React Native FCM workspace and bundle identifier doesn't matter. So we're good here.

mrehan27 (approved):

PR #223, opened 2025-03-06
chore: Enable FCM and APN tests to run on CI

mahmoud-elmorabea:
This PR uses a similar structure for code generation unit tests that we used for the helper native files in the plugin.

The iOS tests are now split into 3 directories:
- Common: includes tests that are not affected by which provider is used APN/FCM
- APN: Tests for generated files with APN specific code
- FCM: Tests for generated files with FCM specific code


Shahroz16 (approved):
lgtm

mrehan27 (approved):

PR #226, opened 2025-03-11
chore: add version verification for customerio-reactnative dependency

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-894/prevent-customers-from-using-incompatible-rn-version-with-expo-plugin

changes:
- Dynamically verify that the installed version matches the peer dependency
- Add warnings at install-time and plugin config-time
- Maintain backward compatibility with existing installations

Implementation:
- Enhanced postInstallHelper.js to validate dependency version
- Added version check in the plugin's entry point
- Improved error messages with specific installation instructions
- Added fallback paths for different installation scenarios

This change is non-intrusive for existing users while helping prevent version mismatch issues going forward.

Screenshot 2025-03-12 at 2 32 35 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,11 +49,12 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": ">=3.4.0"
+    "customerio-reactnative": "4.2.2"
   },
   "devDependencies": {
     "@expo/config-plugins": "^4.1.4",
     "@expo/config-types": "^44.0.0",
+    "@types/fs-extra": "^11.0.4",
     "@typescript-eslint/eslint-plugin": "^5.21.0",
     "@typescript-eslint/parser": "^5.21.0",
     "eslint": "^8.14.0",
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -1,14 +1,32 @@
 import type { ExpoConfig } from '@expo/config-types';
+import * as fs from 'fs-extra';
+import * as path from 'path';
 
 import { withCIOAndroid } from './android/withCIOAndroid';
 import { withCIOIos } from './ios/withCIOIos';
 import type { CustomerIOPluginOptions } from './types/cio-types';
+
+// Get required version from peer dependencies
+function getRequiredSDKVersion() {
+  try {
+    // Use direct path relative to __dirname to be consistent with postInstallHelper
+    const packageJsonPath = path.resolve(__dirname, '../../package.json');
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+    return packageJson.peerDependencies['customerio-reactnative'];
+  } catch (error) {
+    console.warn('Could not determine required SDK version:', error);
+    return null;
+  }
+}
 
 // Entry point for config plugin
 function withCustomerIOPlugin(
   config: ExpoConfig,
   props: CustomerIOPluginOptions
 ) {
+  // Validate dependency during configuration
+  validateDependencies();
+  
   if (props.ios) {
     config = withCIOIos(config, props.ios);
   }
@@ -18,6 +36,32 @@ function withCustomerIOPlugin(
   }
 
   return config;
+}
+
+function validateDependencies() {
+  try {
+    // Load the SDK package.json from node_modules
+    const sdkPath = path.join(process.cwd(), 'node_modules/customerio-reactnative/package.json');
+    const requiredVersion = getRequiredSDKVersion();
+    
+    if (!fs.existsSync(sdkPath)) {
+      console.warn('\x1b[33m%s\x1b[0m', 
+        `Warning: customerio-reactnative is not installed! Please install version ${requiredVersion || 'specified in peer dependencies'}.`
+      );
+      return;
+    }
+    
+    const sdkPkg = JSON.parse(fs.readFileSync(sdkPath, 'utf8'));
+    
+    // Compare versions
+    if (requiredVersion && sdkPkg.version !== requiredVersion) {
+      console.warn('\x1b[33m%s\x1b[0m', 
+        `Warning: customerio-expo-plugin requires customerio-reactnative ${requiredVersion}, but found ${sdkPkg.version}.`
+      );
+    }
+  } catch (error) {
+    console.warn('Could not validate dependencies:', error);
+  }
 }
 
 export default withCustomerIOPlugin;
--- a/plugin/src/postInstallHelper.js
+++ b/plugin/src/postInstallHelper.js
@@ -1,25 +1,88 @@
 const fs = require('fs');
+const path = require('path');
+const findPackageJson = require('find-package-json');
+
+// Get required SDK version from peer dependencies
+function getRequiredSDKVersion() {
+  try {
+    const packageJsonPath = path.resolve(__dirname, '../../package.json');
+    const packageJson = require(packageJsonPath);
+    return packageJson.peerDependencies['customerio-reactnative'];
+  } catch (error) {
+    console.warn('Could not determine required SDK version:', error);
+    return null;
+  }
+}
+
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -49,11 +49,12 @@
     "registry": "https://registry.npmjs.org/"
   },
   "peerDependencies": {
-    "customerio-reactnative": ">=3.4.0"
+    "customerio-reactnative": "4.2.2"
```
mahmoud-elmorabea, line 52: Here's my understanding of how we could solve the issue of having an incompatible version of `customerio-reactnative` when installing our Expo plugin.

I thought that the only thing we were missing is a more strict definition of the supported version.

You can check these articles that explain how `peerDependencies` alone solves this issue.
- [Article 1](https://nodejs.org/en/blog/npm/peer-dependencies)
- [Article 2](https://dev.to/dianjuar/npm-peerdependencies-in-depth-a-comprehensive-introduction-1o6g)
- [Article 3](https://www.solutelabs.com/blog/peer-dependencies-in-npm)

Also a [Firebase example](https://github.com/invertase/react-native-firebase/blob/d9b7175983f0e7b86747c88fcb522295bedf203b/packages/remote-config/package.json#L26) of this

Is that not enough to solve the problem? Have we tried applying this patch (only changing `peerDependencies` `customerio-reactnative` version definition for our v1 Expo plugin and trying to install `customerio-reactnative` v4)

plugin/src/index.ts
```diff
@@ -1,14 +1,32 @@
...
 
 import { withCIOAndroid } from './android/withCIOAndroid';
 import { withCIOIos } from './ios/withCIOIos';
 import type { CustomerIOPluginOptions } from './types/cio-types';
 
+// Get required version from peer dependencies
+function getRequiredSDKVersion() {
+  try {
+    // Use direct path relative to __dirname to be consistent with postInstallHelper
+    const packageJsonPath = path.resolve(__dirname, '../../package.json');
```
mahmoud-elmorabea, line 13: I am curious how safe is it to assume that this is where the customer's app `package.json` will be?

I know that this is a popular setup for Expo apps but for team that can chose to structure their projects differently or for monorepos, this assumption might fail causing our plugin to incorrectly detect mismatch in version that doesn't actually exist.

plugin/src/index.ts
```diff
@@ -20,4 +38,30 @@ function withCustomerIOPlugin(
   return config;
 }
 
-export default withCustomerIOPlugin;
+function validateDependencies() {
+  try {
+    // Load the SDK package.json from node_modules
+    const sdkPath = path.join(process.cwd(), 'node_modules/customerio-reactnative/package.json');
```
mahmoud-elmorabea, line 44: As my previous comment, I think this assumption might not always hold true for all projects structures.

mahmoud-elmorabea (commented):
Thank you so much for the elaborate solution proposal and implementation! 

I am slightly skeptical and concerned that this solution will only work reliably for projects that have standard structure where we can assume where `package.json` and `node_modules` live but if those are different than we expect, we will be failing for valid setup

package.json
Shahroz16: We are still using `peerDependencies` solution while being more strict about it. The rest is just for runtime warnings if there is incompatibility. This was your question right?

plugin/src/index.ts
Shahroz16: We are not loading customers app `pacakge.json` 

```
 // Load the SDK package.json from node_modules
    const sdkPath = path.join(process.cwd(), 'node_modules/customerio-reactnative/package.json');
    const requiredVersion = getRequiredSDKVersion();
 ```
 
 We are looking for the SDK in the node_modules directory

File: plugin/src/index.ts (hunk 3 above)
Comment (Shahroz16) [reply plugin/src/index.ts:44]: okay i get your point now, we are assuming the customerio-reactnative package is hoisted to a same `node_modules` directory, but what if its different, let me look into this case.

File: plugin/src/index.ts [hunk 4]
```diff
@@ -1,14 +1,32 @@
...
 
+// Get required version from peer dependencies
+function getRequiredSDKVersion() {
+  try {
+    // Use direct path relative to __dirname to be consistent with postInstallHelper
+    const packageJsonPath = path.resolve(__dirname, '../../package.json');
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'));
+    return packageJson.peerDependencies['customerio-reactnative'];
+  } catch (error) {
+    console.warn('Could not determine required SDK version:', error);
```
mahmoud-elmorabea, line 17: I don't think we should be printing a warning here since it might be confusing for the customer, this error probably won't be actionable for them

plugin/src/index.ts
```diff
@@ -20,4 +38,31 @@ function withCustomerIOPlugin(
...
 
-export default withCustomerIOPlugin;
+function validateDependencies() {
+  try {
+    // Use Node's own resolution algorithm instead of hardcoded paths
+    let sdkPath;
+    try {
+      sdkPath = require.resolve('customerio-reactnative/package.json');
+    } catch (e) {
+      console.warn('\x1b[33m%s\x1b[0m',
```
mahmoud-elmorabea, line 48: Same as previous comment, can we use this log to only be shown internally? and not for customers? Maybe a debug?

plugin/src/index.ts
```diff
@@ -20,4 +38,31 @@ function withCustomerIOPlugin(
...
+    const requiredVersion = getRequiredSDKVersion();
+    
+    // Compare versions
+    if (requiredVersion && sdkPkg.version !== requiredVersion) {
+      console.warn('\x1b[33m%s\x1b[0m', 
+        `Warning: customerio-expo-plugin requires customerio-reactnative ${requiredVersion}, but found ${sdkPkg.version}.`
+      );
+    }
+  } catch (error) {
+    console.warn('Could not validate dependencies:', error);
```
mahmoud-elmorabea, line 64: Same as previous comment, can we use this log to only be shown internally? and not for customers? Maybe a debug?

plugin/src/postInstallHelper.js
```diff
@@ -1,25 +1,88 @@
...
+const REQUIRED_SDK_VERSION = getRequiredSDKVersion();
+
+function runPostInstall() {
+  // Find the app's root package.json
+  const f = findPackageJson(process.cwd());
+  const appPackageJson = f.next().value;
+  
+  if (!appPackageJson) {
+    console.warn('Could not locate application package.json');
+    return;
```
mahmoud-elmorabea, line 27: Can we test that in case this fails to run, that the npm install will continue to run and won't fail?

plugin/src/index.ts
Shahroz16: All these warnings are for customers? they are supposed to be shown to users when they are adding an incompatible version of the react native SDK. Like this one is telling them exactly what version expo package is using and they should add that as well.

plugin/src/index.ts
Shahroz16: What does showing internally mean here? Only the customer can provide us these logs, so we won't be able to know about them.

plugin/src/postInstallHelper.js
Shahroz16: added them

plugin/src/index.ts
mahmoud-elmorabea: But this warning isn't printed when there is version mismatch, it happens when we can't read version info. So at this point, we are not sure there is a version mismatch or not

plugin/src/index.ts
mahmoud-elmorabea: Internally meaning shown to us during development but not when customers try to install the plugin, not sure if that's possible TBH

mahmoud-elmorabea (approved):
I think it looks good! 

Thanks for adding the tests 😇

PR #227, opened 2025-03-12
chore: handle conflict from firebase 

Shahroz16:
summary:
- Refactors `withGoogleServicesJsonFile.ts` to detect and handle `GoogleService-Info.plist` files across multiple locations
- Prevents duplicate file additions by checking Xcode `project references`
- Improves error handling and logging

changes:
- Systematically checks multiple locations where GoogleService-Info.plist might exist Prioritizes existing files added by React Native Firebase

This PR attempts to resolve build errors caused by both our plugin and React Native Firebase trying to add the same GoogleService-Info.plist file.

### DIFF
--- a/__tests__/ios/fcm/GoogleService-InfoCopied.test.js
+++ b/__tests__/ios/fcm/GoogleService-InfoCopied.test.js
@@ -3,10 +3,46 @@ const path = require("path");
 
 const testProjectPath = path.join(__dirname, "../../../test-app");
 const iosPath = path.join(testProjectPath, "ios");
-const googleServicesFile = path.join(iosPath, "GoogleService-Info.plist");
+const appName = "ExpoTestbed"; // Replace with your app name if different
+// Define all possible locations where GoogleService-Info.plist might exist
+const possibleLocations = [
+  path.join(iosPath, "GoogleService-Info.plist"),                    // Our plugin's default location
+  path.join(iosPath, appName, "GoogleService-Info.plist"),           // Where React Native Firebase typically adds it
+  path.join(iosPath, "Pods", "GoogleService-Info.plist"),            // Sometimes found in Pods
+  path.join(iosPath, appName, "Resources", "GoogleService-Info.plist"), // Alternative location in app resources
+  path.join(iosPath, "Supporting", "GoogleService-Info.plist")       // Another possible location
+];
 
-test("Plugin injects expected customerio-reactnative/apn and customerio-reactnative-richpush/apn in Podfile", async () => {
-  const googleServicesFileExists = fs.existsSync(googleServicesFile);
-
-  expect(googleServicesFileExists).toBe(true);
+describe("GoogleService-Info.plist File", () => {
+  test("GoogleService-Info.plist exists in at least one expected location", () => {
+    // Check all possible locations
+    const existsInAnyLocation = possibleLocations.some(location => fs.existsSync(location));
+    
+    // Log all locations where the file exists
+    const existingLocations = possibleLocations.filter(location => fs.existsSync(location));
+    if (existingLocations.length > 0) {
+      console.log("GoogleService-Info.plist found in these locations:");
+      existingLocations.forEach(location => console.log(`- ${location}`));
+    } else {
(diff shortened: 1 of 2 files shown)


plugin/src/ios/withGoogleServicesJsonFile.ts
```diff
@@ -8,6 +8,167 @@ import { FileManagement } from './../helpers/utils/fileManagement';
...
+ * Checks if a file is already referenced in the Xcode project
+ * 
+ * @param project The Xcode project
+ * @param fileName The file name to check
+ * @returns True if the file is already referenced in the project
+ */
+function isFileReferencedInXcodeProject(project: any, fileName: string): boolean {
+  try {
+    // Get all file references
+    const fileReferences = project.pbxFileReferenceSection();
```
mahmoud-elmorabea, line 21: I have considered this option initially but I was hesitant to do that route since I don't know how reliable it is to be parsing Xcode project file and if this might fail with Xcode upgrades or things like that? Specially since there are no Expo plugins or mods that allow us to access those details in a structured way.

Or maybe even this file looks different for complicated project structures, at the end of the day, the vast majority of iOS development depend on Xcode updating this file when files are added to the project and it's not done manually

__tests__/ios/fcm/GoogleService-InfoCopied.test.js
```diff
@@ -3,10 +3,43 @@ const path = require("path");
...
+      console.log("GoogleService-Info.plist found in these locations:");
+      existingLocations.forEach(location => console.log(`- ${location}`));
+    } else {
+      console.log("GoogleService-Info.plist not found in any expected location");
+    }
+    
+    expect(existsInAnyLocation).toBe(true);
+  });
+  
+  test("GoogleService-Info.plist file is not duplicated", () => {
```
mahmoud-elmorabea, line 30: I think those tests will never fail since our test project doesn't have any Firebase SDKs setup for it, right?

Shahroz16: yes thats correct, created them so when we run these on the branch where you added react native firebase, we can verify it via unit test too. 

also, might help even in future when we want to quickly test any other rn-firebase feature while making sure nothing else breaks. But it can be removed if deemed unnecessary.

plugin/src/ios/withGoogleServicesJsonFile.ts
Shahroz16: I have found the xdode package to be very reliable, the approach is actually the most common way to programmatically interact with Xcode projects in the React Native ecosystem. Most tools (including React Native itself in the CLI), fastlane, expo config use similar methods.

While Apple does change the Xcode project format in xcode upgrades, they generally maintain backward compatibility, especially for core features like file references. 

And lets say there is a case with very complicated structure out of ordinary, and this check fails, the consequences would be we get a duplicated reference which we were going to get anyway? and if someone has this issue we can suggest the solution of not adding the googleservices file in our config?

uros-mil: I've recently updated our Xcode project file for APN UIKit to the newest version (available with Xcode 16.x), and our existing project file parsing (in .github workflow/actions) failed.
I would advise testing this with both Xcode 15.x and 16.x project-file versions before merging.

mahmoud-elmorabea: I actually think we should re-think this change, because a similar logic was problematic with Expo warnings for version conflicts. I believe this is could even be more problematic. 

Like @uros-mil mentions, seems like Xcode 16 added/changed this file format and we don't wanna keep testing and supporting multiple Xcode project files format

uros-mil: I didn't analyze this in depth, but my logic is:
- We'll need a solution that is not error-prone
- If we need to look inside the project file (and there is no other way), let's check if anyone has found an approach that is working well with both newer and older versions of the format

Shahroz16:
Putting this back to draft, so when we plan to bring the expo improvements project again, we can restart the conversation.

PR #233, opened 2025-03-20
chore: revert version validation

Shahroz16:
Version validation was messing up with our local env setup and its not worth the effort to maintain the version validation error warning.


mahmoud-elmorabea (approved):
Thank you 💯

PR #234, opened 2025-03-20
chore: add local script and readme for compatibility scripts

mrehan27:
part of [MBL-975](https://linear.app/customerio/issue/MBL-975/write-scripts-for-test-app-creation-for-different-expo-versions)

### Changes

- Renamed compatibility script prefix from `expo-compat` to `compatibility` based on squad discussion on slack
- Added a README to document compatibility scripts and their usage
- Added a script to run entire compatibility flow using a single command

#### Example Usage

```sh
npm run compatibility:run-compatibility-tests
```


scripts/compatibility/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+```sh
+npm run compatibility:validate-plugin -- --app-path=<path-to-app>
+```
+
+| **Argument** | **Description** | **Default** | **Required** |
+| - | - | - | - |
+| `--app-path` | Path to the test app directory | - | ✅ |
+| `--platforms` | Platforms to test (`android`, `ios`) | `android,ios` | ❌ |
+| `--push-providers` | iOS push providers to test (`apn`, `fcm`) | `apn,fcm` | ❌ |
```
mahmoud-elmorabea, line 66: I am not sure if it makes sense to have separate `--ios-push-provider` and `--push-providers`. From a modularity standpoint, it might make sense.

But when you configure the test app with FCM, you can't really run tests for APN and the other way around. Also, `--push-providers` being a list doesn't make sense because you can never run both FCM and APN on the same app and have them succeed because they are checking for different things at the same files.

Maybe we can only have one flag called `--ios-push-provider` and that is what determines what tests to run.

package.json
```diff
@@ -27,10 +27,11 @@
...
     "test": "jest",
-    "expo-compat:create-test-app": "node scripts/expo-compat/create-test-app.js",
-    "expo-compat:setup-test-app": "node scripts/expo-compat/setup-test-app.js",
-    "expo-compat:configure-plugin": "node scripts/expo-compat/configure-plugin.js",
-    "expo-compat:validate-plugin": "node scripts/expo-compat/validate-plugin.js",
+    "compatibility:create-test-app": "node scripts/compatibility/create-test-app.js",
+    "compatibility:setup-test-app": "node scripts/compatibility/setup-test-app.js",
+    "compatibility:configure-plugin": "node scripts/compatibility/configure-plugin.js",
+    "compatibility:validate-plugin": "node scripts/compatibility/validate-plugin.js",
+    "compatibility:run-compatibility-tests": "node scripts/compatibility/run-compatibility-tests.js",
```
mahmoud-elmorabea, line 34: ✅

mahmoud-elmorabea (changes requested):
This looks great! The readme is very helpful! 

There is one thing though, I tested generating an FCM project, I can see FCM config for CIO plugin but the iOS project has APN generated code.

scripts/compatibility/README.md
mrehan27: Discussed on call. We'll rename the flag to `--ios-push-providers` for clarity and proceed with current implementation.

mahmoud-elmorabea (approved):
Perfect! Thanks 💯

PR #235, opened 2025-03-21
chore: Update unit tests to run on multiple Expo versions

mahmoud-elmorabea:
### Motivation
The strategy we are using to unit test our plugin only using snapshots will not work once we try to run these tests for multiple Expo versions. Different Expo versions generate slightly different native projects which can break snapshot tests.

### Solution
I've used a different technique for some of the tests to allow them the flexibility to run on multiple Expo projects.

#### Gradle files
For Gradle files, we are using `gradle-to-js` to turn Gradle files into a JSON that we can then assert.

#### Android manifest
The manifest is already XML that we can parse into structured data and assert items in it

#### iOS AppDelegate
This is probably the most complicated file we have since it has a lot of code and we inject code within that. I have switched this file to use `inline snapshots` instead where the snapshot data lives in the test file itself. This allows us to inject different app names in the snapshot.

#### Pod files
Pod files are not structured and can't really be parsed, so I used the start/end markets we inject to find the lines to assert.

### Alternatives I've considered
- Duplicating tests for each test app: We could have duplicated tests for each test app with expo version we test, but this would have been too much to maintain
- Regex matching in tests: This gives us the most granular asserting logic but the tests themselves become complex and error prone.


Shahroz16 (approved):
Very happy with the direction we are taking and instead of relying on snapshots we are breaking down problem and solving them as they seem fit. I think we can further enhance it so I have created a follow up [PR](https://github.com/customerio/customerio-expo-plugin/pull/237) which shouldn't block it but would serve as something built on top of it.

package.json
```diff
@@ -126,6 +126,7 @@
   },
   "dependencies": {
     "find-package-json": "^1.2.0",
-    "fs-extra": "^11.2.0"
+    "fs-extra": "^11.2.0",
+    "gradle-to-js": "^2.0.1"
```
mrehan27, line 130: Will this add dependencies to our plugin that will be shipped to customers? Can't we use `devDependencies` or something similar to avoid this?

mrehan27 (approved):
Looks good 👍🏻

PR #239, opened 2025-03-25
chore: Prepare for plugin final release 2.0

mahmoud-elmorabea:
Updating CI scripts to work with `master` branch


.github/workflows/build-sample-apps.yml
```diff
@@ -3,7 +3,7 @@ name: Publish test apps
 on:
   pull_request:
   push:
-    branches: [ beta, feature/* ]
+    branches: [ master, beta, feature/* ]
```
mrehan27, line 6: Can we instead create new `main` branch and use it instead of `master` for consistency across all repos? Or are there any limitations?

.github/workflows/compatibility-tests.yml
mrehan27, line 1: This is already covered by `validate-plugin-compatibility.yml`. I think we can remove it now? 🤔

.github/workflows/build-sample-apps.yml
mahmoud-elmorabea: I am not sure TBH, I'd rather break it down to releasing the final version now, then moving from `master` to `main` separately

Shahroz16 (approved):
I would ➕1 naming it `main` for consistency among repos, but if there is hesitation that it might break the release, I am okay with doing it after.

.github/workflows/build-sample-apps.yml
mrehan27: Makes sense

mrehan27 (approved):

PR #247, opened 2025-04-03
chore: CustomerIOPluginOptionsIOS structure revised for clarity

Ahmed-Ali:
TL;DR:
* All push relevant configuration moved from being properties of `CustomerIOPluginOptionsIOS` to be properties of `CustomerIOPluginOptionsIOS.pushNotification`

--

The `CustomerIOPluginOptionsIOS` type declares the properties use to configure the how the expo plugin will setup CustomerIO SDK.

Part of this is the push notification configurations which are meant to grouped together in the `pushNotification` field for clarity. However, we found out that some push notification configuration flags are placed as fields of the top level type. Moreover, due to this misplacement, some parts of our code expected these flags to be set under the `pushNotification` properties while they aren't and that introduced bugs.

This PR revising this structure to make the object structure easier to reason about.

The impacted properties will be deprecated and will be removed by end of Q2.

## Change Log
In the app.json and app.config.js all the following push notification flags under the `ios`

```json
{
  "ios": {
    "showPushAppInForeground": true,
    "autoTrackPushEvents": true,
    "handleDeeplinkInKilledState": true,
    "disableNotificationRegistration": true,
    "autoFetchDeviceToken": true
  }
}

```
Are move to be part of the `pushNotification` object instead
```json
{
  "ios": {
    "pushNotification": {
      "showPushAppInForeground": true,
      "autoTrackPushEvents": true,
      "handleDeeplinkInKilledState": true,
      "disableNotificationRegistration": true,
      "autoFetchDeviceToken": true
    }
  }
}

```

### DIFF
--- a/__tests__/ios/apn/AppDelegate-impl.test.js
+++ b/__tests__/ios/apn/AppDelegate-impl.test.js
@@ -21,7 +21,7 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <${testAppName()}-Swift.h>
+    #import <ExpoTestbed-Swift.h>
       
     #import "AppDelegate.h"
 
--- a/__tests__/ios/apn/AppDelegate-impl.test.js
+++ b/__tests__/ios/apn/AppDelegate-impl.test.js
@@ -21,7 +21,7 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <ExpoTestbed-Swift.h>
+    #import <${testAppName()}-Swift.h>
       
     #import "AppDelegate.h"
 
@@ -40,6 +40,10 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
       // You can add your custom initial props in the dictionary below.
       // They will be passed down to the ViewController used by React Native.
       self.initialProps = @{};
+
+
+      // Register for push notifications
+      [pnHandlerObj registerPushNotification];
 
       
       [pnHandlerObj initializeCioSdk];
--- a/__tests__/ios/fcm/AppDelegate-impl.test.js
+++ b/__tests__/ios/fcm/AppDelegate-impl.test.js
@@ -20,11 +20,10 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
     #import <EXNotifications/EXNotificationCenterDelegate.h>
     #endif
 
-    @protocol FIRMessagingDelegate;
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <${testAppName()}-Swift.h>
+    #import <ExpoTestbed-Swift.h>
       
     #import "AppDelegate.h"
 
--- a/__tests__/ios/fcm/AppDelegate-impl.test.js
+++ b/__tests__/ios/fcm/AppDelegate-impl.test.js
@@ -20,6 +20,7 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
     #import <EXNotifications/EXNotificationCenterDelegate.h>
     #endif
 
+    @protocol FIRMessagingDelegate;
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
--- a/__tests__/ios/fcm/AppDelegate-impl.test.js
+++ b/__tests__/ios/fcm/AppDelegate-impl.test.js
@@ -23,7 +23,7 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <ExpoTestbed-Swift.h>
+    #import <${testAppName()}-Swift.h>
       
     #import "AppDelegate.h"
 
@@ -42,6 +42,10 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
       // You can add your custom initial props in the dictionary below.
       // They will be passed down to the ViewController used by React Native.
       self.initialProps = @{};
+
+
+      // Register for push notifications
+      [pnHandlerObj registerPushNotification];
 
       
       [pnHandlerObj initializeCioSdk];
--- a/__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
+++ b/__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
@@ -3,7 +3,7 @@
 exports[`Plugin creates expected NotificationService.swift 1`] = `
 "import Foundation
 import UserNotifications
-import CioMessagingPushFCM
+import CioMessagingPushAPN
 
 @objc
 public class NotificationServiceCioManager : NSObject {
@@ -12,7 +12,7 @@ public class NotificationServiceCioManager : NSObject {
   
   @objc(didReceive:withContentHandler:)
   public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-    MessagingPushFCM.initializeForExtension(
+    MessagingPushAPN.initializeForExtension(
       withConfig: MessagingPushConfigBuilder(cdpApiKey: Env.customerIOCdpApiKey)
         .region(Env.customerIORegion)
         .build()
--- a/__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
+++ b/__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
@@ -3,7 +3,7 @@
 exports[`Plugin creates expected NotificationService.swift 1`] = `
 "import Foundation
 import UserNotifications
-import CioMessagingPushAPN
+import CioMessagingPushFCM
 
 @objc
 public class NotificationServiceCioManager : NSObject {
@@ -12,7 +12,7 @@ public class NotificationServiceCioManager : NSObject {
   
   @objc(didReceive:withContentHandler:)
   public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-    MessagingPushAPN.initializeForExtension(
+    MessagingPushFCM.initializeForExtension(
       withConfig: MessagingPushConfigBuilder(cdpApiKey: Env.customerIOCdpApiKey)
         .region(Env.customerIORegion)
         .build()
--- a/plugin/src/ios/withAppDelegateModifications.ts
+++ b/plugin/src/ios/withAppDelegateModifications.ts
@@ -240,19 +240,13 @@ export const withAppDelegateModifications: ConfigPlugin<
       stringContents = addNotificationHandlerDeclaration(stringContents);
 
       // any other value would be treated as true, it has to be explicitly false to disable
-      if (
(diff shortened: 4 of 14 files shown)


plugin/src/ios/withNotificationsXcodeProject.ts
```diff
@@ -299,24 +293,24 @@ const updateNseEnv = (
```
uros-mil, line 293: Let's add early-exit if `options.env` is not object and doesn't exist.
This will make rest of the code easier to read.

We could also deconstruct env to directly use `cdPApiKey` and `region`.

plugin/src/ios/withAppDelegateModifications.ts
```diff
@@ -240,19 +240,13 @@ export const withAppDelegateModifications: ConfigPlugin<
       stringContents = addNotificationHandlerDeclaration(stringContents);
 
       // any other value would be treated as true, it has to be explicitly false to disable
-      if (
-        props.disableNotificationRegistration !== undefined &&
-        props.disableNotificationRegistration === false
-      ) {
+      if (props.pushNotification?.disableNotificationRegistration === false) {
```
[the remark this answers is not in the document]
Shahroz16: +1 to what Uros said, the previous and new check is not the same as previous one. Having said that 

```
props.pushNotification?.disableNotificationRegistration !== true
```

would be different too, 

```
In the original version, the condition is only true when the value exists and is explicitly false.
In the PR version, the condition is only true when the value is explicitly false.
In the uros version, the condition is true for any value except true. This includes undefined, null, false, 0, empty strings, etc.
```

plugin/src/ios/withCIOIos.ts
```diff
@@ -1,21 +1,86 @@
...
+        console.warn(
+          `The ios.${key} property is deprecated. Since the value of ios.pushNotification.${key} is set, it will be used.`
+        );
+      }
+    }
+  });
+
+  if (props.pushNotification?.env !== undefined) {
+    console.warn(
+      'The ios.pushNotification.env property is deprecated. Use ios.nv instead.'
```
mrehan27, line 64:
```suggestion
      'The ios.pushNotification.env property is deprecated. Use ios.env instead.'
```

plugin/src/ios/withNotificationsXcodeProject.ts
```diff
@@ -299,24 +293,24 @@ const updateNseEnv = (
 
   let envFileContent = FileManagement.readFile(envFileName);
 
-  if (options.pushNotification?.env?.cdpApiKey) {
+  if (options.env?.cdpApiKey) {
```
mrehan27, line 296: At the moment, `env` values are only used for NSE when rich push notifications are enabled on iOS. That's why I feel having `env` under `ios.pushNotification` was slightly self explanatory in terms of its purpose. Just wanted to double-check, knowing this, do we still want to move `env` outside of `pushNotification` and directly under `ios` instead? 🤔

Also, just wanted to confirm my understanding. The `mergeDeprecatedPropertiesAndLogWarnings` function above should update properties to always match the new format. Since that always runs first, we only need to read props from new location everywhere else, Correct?

plugin/src/ios/withCIOIos.ts
```diff
@@ -1,21 +1,86 @@
...
+    config = withGoogleServicesJsonFile(config, cioProps);
   }
 
   return config;
 }
+
+const mergeDeprecatedPropertiesAndLogWarnings = (
+  props: CustomerIOPluginOptionsIOS
+) => {
+  const deprecatedTopLevelProperties = {
```
mrehan27, line 30: Not directly related to this code, but adding comment here so we don't forget. Would be nice if we can update compatibility scripts to use newer properties as well (e.g. [`env` here](https://github.com/customerio/customerio-expo-plugin/blob/37415dcc8081cc2a886f347a763f7310ad2a9295/scripts/compatibility/configure-plugin.js#L107-L110)).

mrehan27 (commented):
The compatibility tests are failing. They likely just need script updates to add config properties at new location. That should fix it and we should be good to go.

plugin/src/ios/withAppDelegateModifications.ts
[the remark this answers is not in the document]
Ahmed-Ali: The PR implementation matches exactly the current production implementation and the description on top of the condition.

Seeing these comments makes me realize that this condition is currently a production bug. Changing `===false` to `!== true` should fix it as Uros suggested. I will update the comment too as it is currently incorrect as well.

plugin/src/ios/withCIOIos.ts
```diff
@@ -1,21 +1,86 @@
...
+    autoTrackPushEvents: props.autoTrackPushEvents,
+    handleDeeplinkInKilledState: props.handleDeeplinkInKilledState,
+    disableNotificationRegistration: props.disableNotificationRegistration,
+    autoFetchDeviceToken: props.autoFetchDeviceToken,
+  };
+
+  // loop over all the deprecated properties and log a warning if they are set
+  Object.entries(deprecatedTopLevelProperties).forEach(([key, value]) => {
+    if (value !== undefined) {
+      console.warn(
```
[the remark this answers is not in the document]
Ahmed-Ali: I don't think this is the intention here.
These fields today are at the top level and we are deprecating them in favor of moving them to the `pushNotification` sub object. Hence the warning makes sense to be here. The second warning will only happen if the user set the value in the correct place and in the deprecated field. Which is a second case and its warning seems to also make sense.

plugin/src/ios/withNotificationsXcodeProject.ts
Ahmed-Ali: Ah good point! In that case, they do make sense to be inside push. There is an argument to be made about their best placement, but no strong opinion at this point from my side, hence I will revert that change

plugin/src/ios/withCIOIos.ts
Ahmed-Ali: That makes a lot of sense.
I will add the comments on this one. I also checked the `env` file and since I will revert the `props.pushNotification.env` change, I didn't see anything else there that will be impacted.
I will ping you after I address the rest of the comment to double check with me in case I missed something.

plugin/src/ios/withNotificationsXcodeProject.ts
Ahmed-Ali: I am not sure where the options.env is used in this snippet?
In fact most of the changes here seems to be either: formatting that was missed earlier or my local eslint is kicking in and adding more formatting that we intend.

plugin/src/ios/withNotificationsXcodeProject.ts
```diff
@@ -30,7 +30,7 @@ const addNotificationServiceExtension = async (
       await addPushNotificationFile(options, xcodeProject);
     }
 
-    if (options.pushNotification?.useRichPush) {
+    if (options.pushNotification?.useRichPush === true) {
```
Ahmed-Ali, line 33: This matches our current [documented defaults](https://docs.customer.io/sdk/expo/getting-started/packages-options/#:~:text=ios.pushNotification.useRichPush) better

mrehan27 (commented):
Changes look good. Just two things before we merge:

- We need to update base to `main` since we replaced `master` with `main` recently
- We need to update snapshot tests to ensure CI checks are passing

Rest looks good 👍🏻

__tests__/ios/apn/AppDelegate-impl.test.js
```diff
@@ -21,7 +21,7 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <${testAppName()}-Swift.h>
+    #import <ExpoTestbed-Swift.h>
```
mahmoud-elmorabea, line 24: This should remain dynamic so it works for the base test app as well as the compatibility apps we run tests against different versions of Expo, any reason to change this?

__tests__/ios/fcm/AppDelegate-impl.test.js
```diff
@@ -20,11 +20,10 @@ test('Plugin injects CIO imports and calls into AppDelegate.mm', async () => {
     #import <EXNotifications/EXNotificationCenterDelegate.h>
     #endif
 
-    @protocol FIRMessagingDelegate;
 
     // Add swift bridge imports
     #import <ExpoModulesCore-Swift.h>
-    #import <${testAppName()}-Swift.h>
+    #import <ExpoTestbed-Swift.h>
```
mahmoud-elmorabea, line 26: This should remain dynamic so it works for the base test app as well as the compatibility apps we run tests against different versions of Expo, any reason to change this?

__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Plugin creates expected NotificationService.swift 1`] = `
 "import Foundation
 import UserNotifications
-import CioMessagingPushFCM
+import CioMessagingPushAPN
```
mahmoud-elmorabea, line 6: Looking at the path of the test, this test is specific to FCM and it should import the FCM messaging package

__tests__/ios/fcm/__snapshots__/NotificationService-swift.test.js.snap
```diff
@@ -12,7 +12,7 @@ public class NotificationServiceCioManager : NSObject {
   
   @objc(didReceive:withContentHandler:)
   public func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-    MessagingPushFCM.initializeForExtension(
+    MessagingPushAPN.initializeForExtension(
```
mahmoud-elmorabea, line 15: Looking at the path of the test, this test is specific to FCM and it should import the FCM messaging package

Ahmed-Ali:
Thanks Rehan for the review. The PR is now rebased on top of base. I updated the snapshot tests that's currently working. There are few others who are currently broken in production and worth addressing on their own right rather than trying to squeeze them with this change

plugin/src/ios/withCIOIos.ts
```diff
@@ -10,12 +10,63 @@ export function withCIOIos(
...
+const mergeDeprecatedPropertiesAndLogWarnings = (
+  props: CustomerIOPluginOptionsIOS
+) => {
+  // The deprecatedTopLevelProperties maps the top level properties
+  // that are deprecated to the new ios.pushNotification.* properties
+  // that should be used instead. The deprecated properties are
+  // still available for backwards compatibility, but they will
+  // be removed in the future.
+
+  const deprecatedTopLevelProperties = {
```
mahmoud-elmorabea, line 39: Brilliant! 💯 🥇

mahmoud-elmorabea (commented):
The core change looks great! Absolutely fantastic job @Ahmed-Ali ! I love how we're being considerate by keeping the current behavior and also giving users warnings about deprecation 👍 

One note about the broken tests, I see that the last commit to `main` branch is successfuly, why do we think the tests are broken for production?

I think few minor tweaks to the tests can fix them, I'm okay doing it in this PR or a follow up one 😇

__tests__/ios/fcm/AppDelegate-impl.test.js
Ahmed-Ali: For a second I was like "I never touched these files". It took place when I tried to get jest to help update the snapshot tests `npm run jest -u`. Good learning that we might want to put a hook to fix it on commit

Ahmed-Ali:
> One note about the broken tests, I see that the last commit to main branch is successfuly, why do we think the tests are broken for production?

When I switched to main locally FCM tests kept failing in the same way they fail on this PR. The changes you noted could be the culprit that they were still in place when I checked out main and I didn't pay attention. Will give it another go.

PR #250, opened 2025-04-10
fix: Disable Code Climate Velocity Deployment

mahmoud-elmorabea:
`codeclimate/velocity-deploy-action` Github action started [failing](https://github.com/customerio/customerio-expo-plugin/actions/runs/14379532878/job/40319773344), which is blocking SDK release, we are temporarily disabling it to enables releasing


Shahroz16 (approved):
🚀🚀

PR #253, opened 2025-04-15
chore: Trigger sample apps public builds after npm deployment is successful

mahmoud-elmorabea:
Closes: [MBL-1053](https://linear.app/customerio/issue/MBL-1053/auto-create-testbed-releases-for-every-sdk-release)

This PR triggers a sample apps public release build after the SDK has been deployed successfully to NPM so that internal stakeholders always have up to date versions to test with.


Shahroz16 (approved):
👍

PR #261, opened 2025-05-22
fix: sample apps workflows to run with xcode 16.2

Shahroz16:
changes:

- Updated workflow to utilize the xcode 16.2
- React native SDK version bump


mahmoud-elmorabea (approved):
Looks good! 

I think you can disable the `latest` matrix option for compatibility because it pulls Expo 53 which we don't support yet.

I will be looking into Expo 53 support this week and it will be re-enabled when support is ready

mrehan27 (approved):

PR #265, opened 2025-05-26
chore: Update unit tests and CI scripts to work with Expo 53

mahmoud-elmorabea:
This is a series of PRs that bring Expo 53 support to our plugin.

## PR stack
- Plugin implementation changes -> https://github.com/customerio/customerio-expo-plugin/pull/264
- Tests and CI updates -> **This PR**

## Changes in this PR:
- Updates our compatibility matrix to work with Expo 52 and 53
- Uses Jest's `describe` and `describe.skip` to only run tests relevant to a specific version
- Adds tests for files specific to Expo 53 like the swift version of app delegate

## Testing
- All changes are unit tested


__tests__/utils.js
```diff
@@ -12,4 +12,27 @@ function testAppName() {
...
+    return process.env.EXPO_VERSION || "53.0.0"; // Default to 53
+}
+
+/**
+ * Check if the Expo version is 53 or higher
+ * @returns {boolean} True if Expo version is 53 or higher
+ */
+function isExpoVersion53OrHigher() {
+    const version = getExpoVersion();
+    const majorVersion = parseInt(version.split('.')[0], 10);
```
Ahmed-Ali, line 29: Is there a scenario where the version could be something like `latest` ?

Ahmed-Ali (approved):
It seems like there is quite a bit of duplicate test code between apn and fcm. Since it is unlikely this PR will be a blocker for the actual update, I would say let's give it a pass to reduce the similar test code

.github/workflows/validate-plugin-compatibility.yml
```diff
@@ -110,4 +109,5 @@ jobs:
             --app-path=${{ env.APP_PATH }} \
             --platforms=${{ matrix.platform }}\
             --ios-use-frameworks="static" \
-            ${{ matrix.ios-push-provider && format(' --ios-push-providers={0}', matrix.ios-push-provider) }}
+            ${{ matrix.ios-push-provider && format(' --ios-push-providers={0}', matrix.ios-push-provider) }} \
+            ${{ matrix.expo-version < 53 && format(' --expo-version={0}', matrix.expo-version) }}
```
Shahroz16, line 113: i might be missing something but this might not work? The validation script expects EXPO_VERSION to always be provided, no? and on Expo 53 or latest we wont be providing parameter?

__tests__/ios/apn/__snapshots__/AppDelegate-swift.test.js.snap
```diff
@@ -0,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Expo 53+ AppDelegate tests Plugin injects CIO handler into AppDelegate.swift 1`] = `
+"import Expo
+import React
+import ReactAppDependencyProvider
+
+@UIApplicationMain
+public class AppDelegate: ExpoAppDelegate {
+    let cioSdkHandler = CioSdkAppDelegateHandler()
```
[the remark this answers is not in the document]
Ahmed-Ali: This is a good candidate for a potential follow up revision to our existing lint/eslint rules

__tests__/utils.js
mahmoud-elmorabea: We pass this version through CI, and I only send version numbers and in case of `latest` I just omit it and it defaults to 53, I will change this to use semver as well so the parsing is more robust

__tests__/ios/apn/__snapshots__/AppDelegate-swift.test.js.snap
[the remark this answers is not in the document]
mahmoud-elmorabea: Since this is generated code, we can't/don't really run any lint on it, but maybe we can run the lint on the test snapshots somehow?

PR #280, opened 2025-07-28
chore: Prevent accidental public API changes

mahmoud-elmorabea:
Closes: [MBL-1295](https://linear.app/customerio/issue/MBL-1295/expo-automate-public-api-breaking-changes)

This PR introduces automated API change detection to prevent accidental breaking changes and ensure proper API documentation.

## What's Added
- 📜 API Documentation Script: scripts/generate-api-docs.sh - Generates API reports using Microsoft API Extractor
- 🔧 GitHub Workflow: .github/workflows/check-api-changes.yml - Automatically detects API changes in CI
- ⚡ Package Script: npm run generate-api-docs - Easy local API doc generation with cleanup

## How It Works
- During Development: Run npm run generate-api-docs to update API reports when making intentional API changes
- In CI: Workflow automatically fails if API changes are detected but documentation isn't updated
- Protection: Prevents accidental API breaks by forcing conscious documentation of all public API changes

## Developer Workflow
- Make API changes to TypeScript interfaces
- Run npm run generate-api-docs to update the API report
- Commit both code changes and updated API report
- CI passes ✅
If you forget step 2, CI will fail with clear instructions on how to fix it.

Benefits: Catches breaking changes early, maintains accurate API documentation, enforces deliberate API evolution.


.github/workflows/check-api-changes.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Run API Extractor (Check for API Changes)
```
Shahroz16, line 31: hmm, API Extractor reads the compiled `.d.ts` files under `plugin/libIf`, so if someone edits `plugin/src/**` but forgets to rebuild before committing, CI will still pass because the stale compiled output?

Shahroz16: should we run build or something ourself before checking?

api-extractor.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+  "mainEntryPointFilePath": "<projectFolder>/plugin/lib/typescript/types/cio-types.d.ts",
```
Shahroz16, line 3: `types` is out entry point?

scripts/generate-api-docs.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+source scripts/utils.sh
+set -e
+
+print_heading "Generating API Documentation with API Extractor"
+
+print_blue "\nInstalling dependencies...\n"
+npm install
```
Shahroz16, line 9: if we are calling it here, then ci install is redundant

.github/workflows/check-api-changes.yml
mahmoud-elmorabea: Building TS is part of the process when you run `npm install`. So after you make changes you only need to run `npm run generate-api-docs` and everything will be updated for you

PR #290, opened 2025-08-20
feat: config to override fcm push priority 

Shahroz16:
Problem

When using Customer.io alongside other push notification services (e.g., expo-notifications), both services can register for the same Firebase messaging intent with equal priority, causing conflicts where push notifications may not be delivered to the intended service.

Solution

In all other SDKs, users can override it by updating manifest but in expo we update it

Added `firebaseMessagingServicePriority` configuration option to control Firebase messaging service precedence:

```
  [
    'customerio-expo-plugin',
    {
      android: {
        pushNotification: {
          firebaseMessagingServicePriority: -2  // Lower priority than expo-notifications
        }
      }
    }
  ]
```

  Behavior

  - If specified: Service is added with custom priority regardless of setHighPriorityPushHandler
  - If not specified: Existing behavior is preserved
  - Higher numbers = higher priority, negative numbers = lower priority

  Generated Manifest

```
  <!-- With priority -->
  <service android:name="io.customer.messagingpush.CustomerIOFirebaseMessagingService" android:exported="false">
      <intent-filter android:priority="-2">
          <action android:name="com.google.firebase.MESSAGING_EVENT" />
      </intent-filter>
  </service>
```

  Use Cases

  - Resolve expo-notifications conflict: Set -2 to let expo-notifications handle first
  - Prioritize Customer.io: Set positive number for higher priority
  - No change needed: Existing configs work unchanged

  Type: EnhancementBreaking: No

### DIFF
--- a/__tests__/android/app-manifest.test.js
+++ b/__tests__/android/app-manifest.test.js
@@ -72,3 +72,39 @@ test("Plugin injects notification channel metadata in the app manifest", async (
     expect(channelImportanceMetadata['$']['android:value']).toBe('4');
   });
 });
+
+test("Plugin adds service when firebaseMessagingServicePriority is specified", async () => {
+  // Test that firebaseMessagingServicePriority configuration is supported
+  // Note: This test verifies the service exists. Priority testing would require integration testing
+  // since it depends on the actual plugin compilation process.
+  const manifestContent = await fs.readFile(appManifestPath, "utf8");
+
+  parseString(manifestContent, (err, manifest) => {
+    if (err) throw err;
+
+    const expectedServiceName = 'io.customer.messagingpush.CustomerIOFirebaseMessagingService';
+    const expectedAction = 'com.google.firebase.MESSAGING_EVENT';
+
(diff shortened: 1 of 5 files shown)


mahmoud-elmorabea (changes requested):
I feel like it's confusing to have both `setHighPriorityPushHandler` and `firebaseMessagingServicePriority`. I think we should generally avoid too lower level specific native details in Expo

I suggest by default have the priority to `-2` and only if `setHighPriorityPushHandler` is set to true, we remove the priority so it gets default priority.

This is inline with [our docs ](https://docs.customer.io/integrations/sdk/expo/getting-started/packages-options/#configuring-the-expo-plugin)

> This is optional, if you choose to use a 3rd party plugin to handle notification permissions, but want our SDK to handle the notifications

Maybe we can update the copy a bit, but it would work fine when we're the only push provider and we fall back to let `expo-notification` handle things unless the customer wants CIO to handle push, they can set `setHighPriorityPushHandler` to true.

We can elaborate on that behavior in the multi-push provider section of the docs.

PR #291, opened 2025-08-25
chore: allow auto initialization on ios without push config

mrehan27:
### Summary

Auto initialization previously only worked on iOS when push notifications were configured, which was unintuitive since Android had no such restriction. Additionally, since auto init relies on Swift support (only available in Expo SDK 53+), using it on older versions could silently fail.

### Changes

- Added an explicit validation error if config is used with Expo SDK < 53, preventing confusing or broken builds
- Enabled auto-initialization on iOS even if `pushNotification` is not configured in `app.json`
- Added test cases to validate behavior across both push enabled and disabled configurations

### Testing

- Update test app's `app.json` to remove `pushNotification` block under `ios`
- Run `prebuild` and verify generated `AppDelegate` includes auto init correctly
- iOS app should build and run successfully with full SDK functionality using auto initialization

### DIFF
--- a/plugin/src/ios/withCIOIosSwift.ts
+++ b/plugin/src/ios/withCIOIosSwift.ts
@@ -9,6 +9,7 @@ import { PLATFORM } from '../helpers/constants/common';
 import {
   CIO_CONFIGUREDEEPLINK_KILLEDSTATE_SWIFT_SNIPPET,
   CIO_MESSAGING_PUSH_APP_DELEGATE_INIT_REGEX,
+  CIO_NATIVE_SDK_INITIALIZE_CALL,
   CIO_NATIVE_SDK_INITIALIZE_SNIPPET,
   CIO_REGISTER_PUSHNOTIFICATION_SNIPPET_v2,
   CIO_REGISTER_PUSH_NOTIFICATION_PLACEHOLDER,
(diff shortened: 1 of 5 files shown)


plugin/src/ios/withCIOIosSwift.ts
```diff
@@ -31,10 +32,62 @@ const CIO_SDK_APP_DELEGATE_HANDLER_FILENAME = `${CIO_SDK_APP_DELEGATE_HANDLER_CL
...
+    );
+    return config;
+  }
+
+  // Add files to the Xcode project
+  const xcodeProject = config.modResults;
   const projectRoot = config.modRequest.projectRoot;
   const iosProjectRoot = path.join(projectRoot, 'ios');
+
+  const group = getOrCreateCustomerIOGroup(xcodeProject, projectName);
```
mrehan27, line 51: [`getOrCreateCustomerIOGroup`](https://github.com/customerio/customerio-expo-plugin/blob/9c5c57c2f62cef3c3cc17868e7e978e65a27965e/plugin/src/utils/xcode.ts#L11-L26) is doing the same thing, so no logic change here

plugin/src/ios/withCIOIosSwift.ts
```diff
@@ -201,7 +240,10 @@ const modifyAppDelegate = (
   let modifiedContent = addHandlerPropertyDeclaration(appDelegateContent);
 
   // Modify didFinishLaunchingWithOptions to initialize and call the handler
-  modifiedContent = modifyDidFinishLaunchingWithOptions(modifiedContent);
+  modifiedContent = modifyDidFinishLaunchingWithOptions(
+    modifiedContent,
+    `  cioSdkHandler.application(application, didFinishLaunchingWithOptions: launchOptions)\n\n    `
```
mrehan27, line 245: Copied from `modifyDidFinishLaunchingWithOptions` so it can be reused

Shahroz16 (approved):

PR #297, opened 2025-08-28
chore: improve setHighPriorityPushHandler option

mrehan27:
part of: [MBL-1336](https://linear.app/customerio/issue/MBL-1336/expo-notifications-notificationsaddnotificationreceivedlistener-stops)

### Summary

Improved `setHighPriorityPushHandler` configuration for Android push notification services to better support Expo `prebuild` workflows with other push providers.

### Changes to setHighPriorityPushHandler:

- `undefined` (Default): Unchanged - no service added
- `true`: Unchanged - high priority (no priority attribute)
- `false`: Now always starts at -10 priority (never higher), goes lower if needed

### Other Changes

- Exported `DEFAULT_LOW_PRIORITY = -10` for use across plugin logic and tests
- Ensured consistent behavior during expo `prebuild` so that service priority updates are respected and not skipped
- Added tests to verify that `setHighPriorityPushHandler` applies correct priority value based on the config

#### References

- [Slack thread](https://customerio.slack.com/archives/C06H7MMLH6C/p1756365904925659)

### DIFF
--- a/plugin/src/android/withAndroidManifestUpdates.ts
+++ b/plugin/src/android/withAndroidManifestUpdates.ts
@@ -4,9 +4,37 @@ import type { ManifestApplication } from '@expo/config-plugins/build/android/Man
 
 import type { CustomerIOPluginOptionsAndroid } from '../types/cio-types';
 
+// Default low priority for Firebase messaging service when setHighPriorityPushHandler is false
+export const DEFAULT_LOW_PRIORITY = -10;
+
+// Helper function to calculate target priority for low priority services
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function calculateLowPriority(services: any[], excludeIndex?: number): number {
+  const relevantServices = excludeIndex !== undefined
+    ? services.filter((_, index) => index !== excludeIndex)
+    : services;
+
+  const existingPriorities = relevantServices
+    .flatMap(service => service['intent-filter'] || [])
+    .map(filter => (filter.$ as Record<string, string>)?.['android:priority'])
+    .filter(priority => priority !== undefined)
+    .map(priority => parseInt(priority, 10))
+    .filter(priority => !isNaN(priority));
+
+  let targetPriority = DEFAULT_LOW_PRIORITY;
+  if (existingPriorities.length > 0) {
+    const minExistingPriority = Math.min(...existingPriorities);
+    if (minExistingPriority <= DEFAULT_LOW_PRIORITY) {
+      targetPriority = minExistingPriority - 1;
+    }
+  }
+
+  return targetPriority;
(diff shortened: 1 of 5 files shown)


plugin/src/android/withAndroidManifestUpdates.ts
```diff
@@ -17,31 +45,81 @@ export const withAndroidManifestUpdates: ConfigPlugin<
...
-            ],
-          },
-        ],
+        'intent-filter': [intentFilter],
       });
-      console.log(
-        'Successfully set CustomerIO push handler as priority in AndroidManifest.xml'
-      );
+    } else if (options.setHighPriorityPushHandler === true) {
+      // Service exists, need to ensure it becomes high priority (remove priority attribute)
```
mahmoud-elmorabea, line 91: What is the scenarios where the service would already exists? Only case if customer is checking in their native projects in source code and then do we want to override something that is added manually?

plugin/src/android/withAndroidManifestUpdates.ts
```diff
@@ -17,31 +45,81 @@ export const withAndroidManifestUpdates: ConfigPlugin<
...
+            'Successfully updated existing CustomerIO push handler to high priority in AndroidManifest.xml'
+          );
+        }
+      }
+    } else if (options.setHighPriorityPushHandler === false) {
+      // Service exists, but we need to update its priority when flag is false
+      const existingService = application[0].service[existingServiceIndex];
+
+      // Calculate target priority excluding the existing CustomerIO service
+      const targetPriority = calculateLowPriority(application[0].service, existingServiceIndex);
```
mahmoud-elmorabea, line 109: I don't think we should be doing that:
- It overrides something the customer might have added or configured
- This isn't the merged manifest, there could be other services added by dependencies that won't be accounted for here

mahmoud-elmorabea (commented):
Good plan, I think we should just simplify things:
- high priority == true -> Add service with default priority
- high priority == false -> Add service with negative priority (magic number to best of our guess)

plugin/src/android/withAndroidManifestUpdates.ts
mrehan27: Running `prebuild` again without `--clean` flag

plugin/src/android/withAndroidManifestUpdates.ts
mrehan27: Responded in slack thread we discussed earlier for quicker resolution.

mrehan27: Removed it as discussed on slack

mahmoud-elmorabea (approved):

PR #300, opened 2025-09-05
chore: support only config without notification setup

wcastand:
I didn't do much since most of the code was already handling the possibility, just remove the props.ios and props.android that was preventing the modification to happen without the notification config.

made the function accept undefined notification config.

i tried to build the test-app with only `config` in the plugin and everything seemed to work. 

I started looking into making an test-app with no notification setup and test but that was making a massive PR so figured, before spending time on that, let's see what you think and if you want and how you'd like that to be tested :)

### DIFF
--- a/plugin/src/android/withCIOAndroid.ts
+++ b/plugin/src/android/withCIOAndroid.ts
@@ -15,8 +15,6 @@ export function withCIOAndroid(
   sdkConfig: NativeSDKConfig | undefined,
   props?: CustomerIOPluginOptionsAndroid
 ): ExpoConfig {
-
-    config = withProjectStrings(config);
   // Only run notification setup if props are provided
   if(props) {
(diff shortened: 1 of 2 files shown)


mrehan27 (commented):
Can you please resolve the comment, update with `main`, and ensure `lint` is passing? I have a similar PR ready to push this and other changes live. If you don't get time to update, let me know so I can close this and open mine.

wcastand:
should be good to do then, move projectStrings outside of the `if` and fix conflicts

mrehan27 (commented):
Minor comments only, rest looks good

wcastand:
made the small changes 👍

mrehan27:
I’ll merge this as `chore` once CI passes and push the release after final testing. Thanks for the changes. Please keep them coming 🙌🏻

PR #334, opened 2026-04-27
chore: add pnpm dev apps + CI to reproduce monorepo build issue

mahmoud-elmorabea:
## Summary

- Adds two minimal Expo dev apps that reproduce the build failures customers hit when integrating the plugin in pnpm and pnpm-workspace layouts: `test-app-pnpm/` (flat) and `test-app-pnpm-monorepo/` (workspace where `apps/mobile` shares `customerio-reactnative` with `packages/shared-cio-utils`, exercising pnpm's deduplication and symlinking).
- Adds dedicated convenience commands and setup scripts (`setup-test-app-pnpm`, `setup-test-app-pnpm-monorepo`, `cleanAllPnpm`, `buildAllPnpm`, `cleanAndBuildAllPnpm`). The existing `cleanAll` / `buildAll` defaults stay focused on the npm `test-app/`.
- Adds a dedicated CI workflow (`.github/workflows/test-pnpm-compatibility.yml`) on `macos-15` with two sequential phases against the new dev apps:
  1. **Isolated mode** (default pnpm) — runs both setup scripts (which invoke `expo prebuild --clean` → triggers `pod install`) and asserts:
     - The generated `Podfile` does **not** contain `.pnpm/` on any `:path =>` line.
     - `expoVersion` is present in the installed `customerio-reactnative/package.json` (otherwise iOS User-Agent attribution silently reports `ReactNative` instead of `Expo`).
  2. **Hoisted mode** — wipes the monorepo install, writes the recommended `.npmrc` (`node-linker=hoisted`, `shamefully-hoist=true`, `public-hoist-pattern[]=*customerio*`), re-runs setup, and re-runs the same assertions. Guards the install path we point customers at when they need an immediate workaround.

- Adds an `eas-build-post-install` hook to both dev apps (`test-app-pnpm/package.json` and `test-app-pnpm-monorepo/apps/mobile/package.json`) so manual `eas build` runs exercise the same prebuild path EAS uses in production. Every real-world Expo monorepo we sampled defines this hook; without it, the dev apps wouldn't cover that code path.

## Context

A reported build failure surfaced `pod install` rejecting the install with `[!] There are multiple dependencies with different sources for "customerio-reactnative" in "Podfile"`, which downstream surfaced as a `NativeCustomerIO` TurboModule runtime error.

Root cause is in the plugin's path resolver — under pnpm, `node_modules/customerio-reactnative` is a symlink and the resolver follows it through `realpath` to `node_modules/.pnpm/customerio-reactnative@HASH/...`. That long path gets written into the Podfile while React Native autolinking writes the symlink path, and CocoaPods rejects the disagreement. The hash also changes whenever the dependency tree changes, so the emitted path is non-deterministic across installs.

Separately, the postinstall hook that writes `expoVersion` into the RN SDK's `package.json` (consumed by `customerio-cdp.ts:43-46` → `NativeCustomerIO.swift:42-46` → `UserAgentUtil.swift:31` to set the iOS User-Agent header) is silently skipped under pnpm, so every Expo customer's iOS API requests today report `User-Agent: Customer.io ReactNative Client/<rn-version>` instead of `Expo Client/<plugin-version>`.

## Stacked PR note

This PR is **the reproduction half** of a two-PR stack against `feature/pnpm-support`:

1. **This PR** — adds the dev apps + verification CI. The pnpm-compatibility workflow is **intentionally red on this branch** because the plugin's resolver still emits `.pnpm/...` paths and `expoVersion` never lands. That's the regression demo.
2. **Follow-up PR** — adds the resolver and postinstall fixes plus Jest fixture tests for the path resolver. Once stacked on top, the same CI workflow goes green.

The red→green progression across the two PRs is the end-to-end demonstration that the fix works against the bug.

---

> [!NOTE]
> **Medium Risk**
> Adds new CI coverage and substantial new test-app fixtures/scripts; low impact on shipped plugin code but moderate risk of CI flakiness/time increases due to macOS `expo prebuild`/CocoaPods runs.
> 
> **Overview**
> Adds two new minimal Expo dev apps (`test-app-pnpm` and `test-app-pnpm-monorepo`) to exercise pnpm isolated installs and pnpm workspaces (including a shared package to force pnpm dedupe/symlinks), along with new root scripts to set up, clean, and build them via a packaged plugin tarball.
> 
> Introduces a macOS GitHub Actions workflow (`test-pnpm-compatibility.yml`) that runs these setups in both default pnpm and *hoisted* modes, then verifies the generated iOS `Podfile` doesn’t contain `.pnpm/` `:path` entries and that `expoVersion` is present in the installed `customerio-reactnative` `package.json`. Documentation and TS configs are updated to exclude the new dev app directories from builds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddd42c4e602efe0f578b3849eee8474be4fb5cde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mahmoud-elmorabea:
The failing CI job proves that without the fix, the build for mono-repo pnpm fails. Next PR fixes the same compatibility test https://github.com/customerio/customerio-expo-plugin/pull/335

.github/workflows/test-pnpm-compatibility.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: 'Phase A — set up test-app-pnpm-monorepo (workspace) — includes pod install'
+        run: npm run setup-test-app-pnpm-monorepo
+
+      - name: 'Phase A — verify Podfile :path values and expoVersion attribution'
+        run: npm run verify-pnpm-dev-apps
+
+      # Phase B — hoisted mode.
+      # Customers who can't wait for a release are advised to add a
+      # node-linker=hoisted .npmrc as a workaround. This phase exercises
```
mrehan27, line 45: Does this need to be added in expo doc?

mrehan27 (approved):

.github/workflows/test-pnpm-compatibility.yml
mahmoud-elmorabea: Not really, this was our advice/workaround before this support. We don't want to suggest that anymore

PR #337, opened 2026-04-29
chore: inject killed-state deep link workaround before factory.startReactNative

mahmoud-elmorabea:
## Summary

- On modern Expo Swift `AppDelegate.swift` templates, RN is bootstrapped by `factory.startReactNative(...)` *inside* the `#if os(iOS) || os(tvOS)` guard — *before* the trailing `return super.application(...)`. The previous injection placed the deep-link extraction block before the return statement, so on the new template it ran after RN had already been initialized, making `handleDeeplinkInKilledState` a silent no-op.
- Inject the snippet before the `#if` guard and route `modifiedLaunchOptions` into `factory.startReactNative` so RN sees the URL on cold start. The trailing `super.application(...)` is still rewritten to consume `modifiedLaunchOptions` for backward compatibility with older templates where `super` was the call that bootstrapped RN.
- Backfill the previously-untested option with positional regression tests, a snapshot, and prebuild snapshot updates so this regression can't recur silently.

## Why this slipped through

The `handleDeeplinkInKilledState` flag had **zero** automated coverage on either the Swift or Objective-C path — `__tests__/` contained no references to it. When the Expo Swift template moved RN bootstrap from `super.application(...)` to `factory.startReactNative(...)`, no snapshot diffed and no assertion fired, so the workaround quietly became a no-op without anyone noticing. (The deep link still routes correctly today via the CIO native SDK's `UNUserNotificationCenterDelegate` click handler — that path runs regardless of this option — but the flag itself was no longer doing anything.)

## What changed

| File | Change |
| --- | --- |
| `plugin/src/ios/withCIOIosSwift.ts` | `addHandleDeeplinkInKilledState` now detects modern templates (with `factory.startReactNative` inside an `#if os(iOS) \|\| os(tvOS)` guard), inserts the snippet before that guard, and rewrites the factory call's `launchOptions:` argument. Falls back to the previous behavior for legacy templates without `factory.startReactNative`. |
| `__tests__/ios/withCIOIosSwift.test.ts` | New `handleDeeplinkInKilledState placement` describe block: 4 tests covering position invariants, the `modifiedLaunchOptions` rewrite, the legacy fallback path, and idempotency. Two also snapshot the full transformed AppDelegate to surface content drift in the injected snippet itself. |
| `__tests__/ios/__snapshots__/withCIOIosSwift.test.ts.snap` | New (auto-generated). |
| `__tests__/ios/{apn,fcm}/__snapshots__/AppDelegate-swift.test.js.snap` | Updated to reflect the prebuild output now that the flag is enabled in `test-app/app.json`. |
| `test-app/app.json` | Set `pushNotification.handleDeeplinkInKilledState: true` so the prebuild-driven snapshot tests cover this code path going forward. |
| `.github/workflows/test.yml` | Wire `withCIOIos.test.ts` and `withCIOIosSwift.test.ts` into the `test-utils` job — they were previously orphaned (not matched by the `test-plugin-ios` path filter) and silently not running. |

## Test plan

- [x] `npx jest __tests__/utils __tests__/ios/withCIOIos.test.ts __tests__/ios/withCIOIosSwift.test.ts` — 7 suites, 111 tests, 6 snapshots, all green locally. This is the exact command the `test-utils` CI job will run.
- [x] Snapshot diffs eyeballed: `factory.startReactNative` consumes `modifiedLaunchOptions`; deep-link block sits before `#if os(iOS) \|\| os(tvOS)`; trailing `super.application(...)` also rewritten.
- [ ] CI `test-plugin-ios` (apn + fcm matrix) — should pass now that the APN/FCM `AppDelegate-swift` snapshots reflect the post-fix output. If a CI run regenerates a slightly different prebuild artifact (e.g., Expo template formatting drift), the snapshot diff will be visible in the failing job and easy to update.
- [ ] CI `test-utils` — should pick up the 4 new positional/snapshot tests + 7 existing tests in `withCIOIosSwift.test.ts` and the existing `withCIOIos.test.ts` cases.
- [ ] Cold-start manual verification on a physical device with `handleDeeplinkInKilledState: true`: tap a CIO push containing a deep link from killed state, confirm `Linking.getInitialURL()` returns the URL (Path 1 functioning) rather than `null` (Path 2 fallback only).

## Notes

- No changes to `plugin/src/ios/withAppDelegateModifications.ts` (Objective-C path) — verified that path already routes `modifiedLaunchOptions` into `createBridgeWithDelegate:launchOptions:` correctly.
- No changes to `plugin/src/helpers/constants/ios.ts` — the snippet itself is reused as-is.

---

> [!NOTE]
> **Medium Risk**
> Changes native code generation for iOS `AppDelegate.swift` via regex-based injection; incorrect matching could break app startup, but the change is scoped and now covered by targeted snapshots/tests.
> 
> **Overview**
> Fixes the iOS Swift `handleDeeplinkInKilledState` injection so it works on modern Expo templates by inserting the deep-link workaround *before* the `#if os(iOS) || os(tvOS)` block and rewriting `factory.startReactNative(... launchOptions:)` (and the trailing `super.application(...)`) to use `modifiedLaunchOptions`.
> 
> Adds regression coverage with new Swift placement/idempotency tests plus snapshots, updates APN/FCM AppDelegate snapshots accordingly, and enables `handleDeeplinkInKilledState: true` in the sample app/default compatibility config. CI `test-utils` is updated to actually run the previously-orphaned `__tests__/ios/withCIOIos*.test.ts` suites.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8dcb051251c86716fbe7ffa613036dd5c41fb13. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,4 +76,4 @@ jobs:
         run: npm ci
 
       - name: Run utils tests
-        run: npm test -- __tests__/utils
+        run: npm test -- __tests__/utils __tests__/ios/withCIOIos.test.ts __tests__/ios/withCIOIosSwift.test.ts
--- a/__tests__/ios/__snapshots__/withCIOIosSwift.test.ts.snap
+++ b/__tests__/ios/__snapshots__/withCIOIosSwift.test.ts.snap
@@ -1,0 +1,141 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`withCIOIosSwift handleDeeplinkInKilledState placement injects deeplink workaround BEFORE factory.startReactNative on modern Expo Swift templates 1`] = `
+"import Expo
+import React
+import ReactAppDependencyProvider
+
+@UIApplicationMain
+public class AppDelegate: ExpoAppDelegate {
+  let cioSdkHandler = CioSdkAppDelegateHandler()
+
+  var window: UIWindow?
+
+  var reactNativeDelegate: ExpoReactNativeFactoryDelegate?
+  var reactNativeFactory: RCTReactNativeFactory?
+
+  public override func application(
+    _ application: UIApplication,
+    didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
+  ) -> Bool {
+    let delegate = ReactNativeDelegate()
+    let factory = ExpoReactNativeFactory(delegate: delegate)
+    delegate.dependencyProvider = RCTAppDependencyProvider()
+
+    reactNativeDelegate = delegate
+    reactNativeFactory = factory
+    bindReactNativeFactory(factory)
+
+    // Deep link workaround for app killed state start
+    var modifiedLaunchOptions = launchOptions
+    if let launchOptions = launchOptions,
+       let pushContent = launchOptions[UIApplication.LaunchOptionsKey.remoteNotification] as? [AnyHashable: Any],
+       let cio = pushContent["CIO"] as? [String: Any],
+       let push = cio["push"] as? [String: Any],
+       let link = push["link"] as? String,
+       !launchOptions.keys.contains(UIApplication.LaunchOptionsKey.url) {
+        
+        var mutableLaunchOptions = launchOptions
+        mutableLaunchOptions[UIApplication.LaunchOptionsKey.url] = URL(string: link)
+        modifiedLaunchOptions = mutableLaunchOptions
+    }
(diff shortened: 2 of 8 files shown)


mrehan27 (approved):
Looks good overall. I believe we already have, but just wanted to double check that the tests now cover all paths for `handleDeeplinkInKilledState` config (i.e., `true`/`false`/`undefined`), with the default value unchanged.

PR #359, opened 2026-05-11
fix(ios): correct PBXContainerItemProxy bucket aliasing (MBL-1710)

### LINKED ISSUE #350: withNotificationsXcodeProject.js: typo causes orphaned PBXContainerItemProxy references — breaks @bacons/xcode strict serializers (Expo SDK 55+)
There's a one-character typo in plugin/src/ios/withNotificationsXcodeProject.ts (and the compiled outputs in                                              
  plugin/lib/{commonjs,module}/ios/withNotificationsXcodeProject.js) that's been present at least since 2.11.0 and is still in 3.3.0.
                                                                                                                                                            
## The bug                                                                                                                                                
                  
  `plugin/src/ios/withNotificationsXcodeProject.ts:196`:                                                                                                    
                  
  ```js                                                                                                                                                     
  projObjects.PBXTargetDependency = projObjects.PBXTargetDependency || {};
  projObjects.PBXContainerItemProxy = projObjects.PBXTargetDependency || {};
  ```                                                                                                                                                       
   
  ## Fix                                                                                                                                                    
                  
  ```diff
  - projObjects.PBXContainerItemProxy = projObjects.PBXTargetDependency || {};
  + projObjects.PBXContainerItemProxy = projObjects.PBXContainerItemProxy || {};                                                                            
  ```       
                                                                                                                                                            
  The second line should reference PBXContainerItemProxy, not PBXTargetDependency. As written, it aliases PBXContainerItemProxy to the same object reference as PBXTargetDependency (both {} after init). When xcodeProject.addTarget() then writes a target dependency and its companion proxy, both writes land in a shared object — the proxy entries effectively land in the wrong bucket / get overwritten by dependency entries.                                          
                  
  Fix

  -projObjects.PBXContainerItemProxy = projObjects.PBXTargetDependency || {};
  +projObjects.PBXContainerItemProxy = projObjects.PBXContainerItemProxy || {};                                                                            
                                                                                                                                                            
Your existing workaround code for addTarget's legacy bug contains a one-character typo on the second line. Historically harmless because the legacy xcode serializer tolerated the resulting orphan. Expo SDK 55's new strict @bacons/xcode serializer doesn't — so any user combining CustomerIO with another plugin that adds Xcode targets (like @bacons/apple-targets, expo-app-blocker, react-native-device-activity) on SDK 55+ hits a hard crash.                                                                                                                                              
                  
                                                                                                                                                             
  ```                                                                                                                                                       
  [Malformed Xcode project]: Found orphaned reference: <UUID> > PBXTargetDependency.targetProxy > <UUID>                                                    
  Error: [ios.xcodeProjectBeta2]: withIosXcodeProjectBeta2BaseMod: Unable to serialize object: '<UUID>'
  ```                                                                                                                                                       
   
  ```                                                                                                                                                       
  TypeError: [ios.xcodeProjectBeta2]: withIosXcodeProjectBeta2BaseMod: dep.props.targetProxy.isRemote is not a function
      at PBXTargetDependency.toJSON (/.../@bacons/xcode/build/api/AbstractObject.js:130:27)
  ```

  Specifically affected: any consumer using Expo SDK 55+ alongside another plugin that adds Xcode targets (e.g., react-native-device-activity,
  expo-app-blocker, @bacons/apple-targets directly, expo-share-extension, expo-widgets in some configs).

  Repro

  1. Expo SDK 55 app
  2. Install customerio-expo-plugin (any version through 3.3.0) and any plugin that registers @bacons/apple-targets (e.g., expo-app-blocker)
  3. npx expo prebuild --platform ios --clean
  4. Crashes with one of the two stack traces above.

  Workaround

  pnpm patch customerio-expo-plugin and apply the one-line fix to all three locations:

  - plugin/src/ios/withNotificationsXcodeProject.ts (line 196 in 2.11.0)
  - plugin/lib/commonjs/ios/withNotificationsXcodeProject.js (line 154 in 2.11.0)
  - plugin/lib/module/ios/withNotificationsXcodeProject.js (line 147 in 2.11.0)

mrehan27:
## Summary

- Fixes a one-character typo in `addNotificationServiceExtensionToXcodeProject` that aliased `PBXContainerItemProxy` to the same object reference as `PBXTargetDependency`. After `addTarget()` ran, both buckets shared one object, leaving the `PBXContainerItemProxy` key pointing at records with `isa: PBXTargetDependency`.
- Expo SDK ≤54 (legacy `xcode` serializer) tolerated the orphan; Expo SDK 55+ strict `@bacons/xcode` serializer crashes when walking `PBXTargetDependency.targetProxy` UUIDs, so any consumer who composes us with another plugin that adds Xcode targets (`@bacons/apple-targets`, `expo-app-blocker`, `react-native-device-activity`, `expo-share-extension`, some `expo-widgets` setups) hits a hard `expo prebuild` failure.
- Adds a regression test that (a) asserts `PBXTargetDependency` and `PBXContainerItemProxy` are distinct object references after the helper runs, and (b) walks every `PBXTargetDependency.targetProxy` UUID and asserts it resolves to a record with `isa === 'PBXContainerItemProxy'`. Verified the test fails against the buggy code and passes against the fix.

Bug existed since at least 2.11.0; reported externally in #350.

## Links

- Linear: [MBL-1710](https://linear.app/customerio/issue/MBL-1710/withnotificationsxcodeprojectjs-typo-causes-orphaned)
- Closes #350

## Test plan

- [x] `npx jest __tests__/scenarios/ios/notificationServiceXcodeProject.test.ts` — 9/9 pass, including the new regression test
- [x] Verified regression test fails on the buggy code (the failure output shows `PBXTargetDependency` and `PBXContainerItemProxy` entries intermingled in the shared bucket) and passes on the fix
- [x] `npm run typescript` — clean
- [x] `npm run lint` — clean (one pre-existing unrelated warning in `helpers/constants/ios.ts`)
- [x] **End-to-end on Expo SDK 55** — scratch app (`expo ~55.0.23`, `@expo/config-plugins 55.0.8`, `@bacons/xcode 1.0.0-alpha.32`) with `customerio-expo-plugin` (`useRichPush: true`) plus `expo-app-blocker` and `@bacons/apple-targets` widget — `npx expo prebuild --platform ios --clean` succeeds; generated pbxproj has correctly-typed entries (5 `PBXContainerItemProxy` + 5 `PBXTargetDependency`, all properly isa-tagged).
- [x] **On-disk impact of the bug** confirmed by reverting the fix in the installed copy and re-prebuilding: with the bug, the generated pbxproj contains duplicate UUID entries across both sections (2 `PBXContainerItemProxy` + 2 `PBXTargetDependency` for a project with a single NSE target) — exactly the structural defect that the strict serializer rejects. With the fix applied: 1 of each, clean shape.

---

> [!NOTE]
> **Medium Risk**
> Touches iOS Xcode project mutation during target creation; a small change but it affects pbxproj structure and can break prebuild/serialization if incorrect.
> 
> **Overview**
> Fixes a typo in `addNotificationServiceExtensionToXcodeProject` so `PBXContainerItemProxy` is initialized as its own section instead of aliasing `PBXTargetDependency`, preventing mixed/invalid records after `addTarget()`.
> 
> Adds a regression test asserting the `PBXTargetDependency` and `PBXContainerItemProxy` buckets contain only records with the correct `isa`, guarding against future pbxproj structure corruption (notably with strict serializers in newer Expo toolchains).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5995dc438bb087677291863861f4899373132e29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
+++ b/__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
@@ -150,49 +150,35 @@ describe('ios scenarios — addNotificationServiceExtensionToXcodeProject (vanil
     }
   });
 
-  // Regression for #350: the legacy-bug workaround initialized
-  // PBXTargetDependency and PBXContainerItemProxy with the SAME object
-  // reference (a one-character typo aliased proxy to the dependency bucket).
-  // The legacy serializer tolerated the resulting orphan proxy entries,
-  // but Expo SDK 55+ strict @bacons/xcode serializer crashes when it walks
-  // the project and finds a PBXTargetDependency.targetProxy UUID that has
-  // no entry in PBXContainerItemProxy.
-  it('initializes PBXTargetDependency and PBXContainerItemProxy as distinct objects, so addTarget populates both buckets independently', () => {
+  // Regression for #350: a one-character typo aliased PBXContainerItemProxy
+  // to the same object reference as PBXTargetDependency, so addTarget's
+  // dependency and proxy records both landed in a shared bucket. The legacy
+  // serializer tolerated the orphan; Expo SDK 55+ strict @bacons/xcode
+  // serializer crashes. The invariant: each bucket holds records of its own
+  // isa only. If the buckets were aliased again, both assertions below fail.
+  it('populates PBXTargetDependency and PBXContainerItemProxy with well-typed records', () => {
     const { project } = loadFixture('vanilla.pbxproj');
     addNotificationServiceExtensionToXcodeProject(project, baseOptions);
 
     const objects = project.hash.project.objects;
-    const dependency = objects.PBXTargetDependency;
-    const proxy = objects.PBXContainerItemProxy;
-
-    expect(dependency).toBeDefined();
-    expect(proxy).toBeDefined();
-    // If aliased, the two buckets share an object reference and every
-    // proxy entry shows up as a dependency entry (and vice versa).
-    expect(proxy).not.toBe(dependency);
-
-    // Every PBXTargetDependency entry references a UUID via targetProxy;
-    // that UUID must resolve inside PBXContainerItemProxy. With the typo,
-    // the proxy bucket is the same object as the dependency bucket, so
-    // the lookup finds a PBXTargetDependency record (which has no isa of
-    // PBXContainerItemProxy) and the strict serializer rejects it.
-    const dependencyEntries = Object.entries(dependency as Record<string, unknown>)
-      .filter(([key]) => !key.endsWith('_comment'))
-      .map(([, value]) => value)
-      .filter(
-        (value): value is { isa: string; targetProxy: string } =>
-          typeof value === 'object' &&
-          value !== null &&
-          (value as { isa?: string }).isa === 'PBXTargetDependency',
-      );
-    expect(dependencyEntries.length).toBeGreaterThan(0);
-    for (const entry of dependencyEntries) {
-      const referenced = (proxy as Record<string, unknown>)[entry.targetProxy];
-      expect(referenced).toBeDefined();
-      expect((referenced as { isa: string }).isa).toEqual('PBXContainerItemProxy');
-    }
+    assertBucketIsHomogeneous(objects.PBXTargetDependency, 'PBXTargetDependency');
+    assertBucketIsHomogeneous(objects.PBXContainerItemProxy, 'PBXContainerItemProxy');
   });
 });
+
+const assertBucketIsHomogeneous = (
+  bucket: unknown,
+  expectedIsa: string,
+): void => {
+  expect(bucket).toBeDefined();
+  const entries = Object.entries(bucket as Record<string, unknown>).filter(
+    ([key]) => !key.endsWith('_comment'),
+  );
+  expect(entries.length).toBeGreaterThan(0);
+  for (const [, value] of entries) {
+    expect(value).toMatchObject({ isa: expectedIsa });
+  }
+};
 
 // Behavior finding (do NOT change in this refactor): the helper's early-exit
 // guard `if (xcodeProject.pbxTargetByName(CIO_NOTIFICATION_TARGET_NAME))` was
--- a/__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
+++ b/__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
@@ -149,6 +149,49 @@ describe('ios scenarios — addNotificationServiceExtensionToXcodeProject (vanil
       );
     }
   });
+
+  // Regression for #350: the legacy-bug workaround initialized
+  // PBXTargetDependency and PBXContainerItemProxy with the SAME object
+  // reference (a one-character typo aliased proxy to the dependency bucket).
+  // The legacy serializer tolerated the resulting orphan proxy entries,
+  // but Expo SDK 55+ strict @bacons/xcode serializer crashes when it walks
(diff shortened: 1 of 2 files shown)


__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
```diff
@@ -149,6 +149,49 @@ describe('ios scenarios — addNotificationServiceExtensionToXcodeProject (vanil
...
+          value !== null &&
+          (value as { isa?: string }).isa === 'PBXTargetDependency',
+      );
+    expect(dependencyEntries.length).toBeGreaterThan(0);
+    for (const entry of dependencyEntries) {
+      const referenced = (proxy as Record<string, unknown>)[entry.targetProxy];
+      expect(referenced).toBeDefined();
+      expect((referenced as { isa: string }).isa).toEqual('PBXContainerItemProxy');
+    }
+  });
```
mahmoud-elmorabea, line 166: How about we do something like this instead?

```
  const assertBucketIsHomogeneous = (
    bucket: unknown,
    expectedIsa: string,
  ): void => {
    expect(bucket).toBeDefined();
    const entries = Object.entries(bucket as Record<string, unknown>).filter(
      ([key]) => !key.endsWith('_comment'),
    );
    expect(entries.length).toBeGreaterThan(0);
    for (const [, value] of entries) {
      expect(value).toMatchObject({ isa: expectedIsa });
    }
  };

  it('populates PBXTargetDependency and PBXContainerItemProxy with well-typed records (regression for #350)', () => {
    const { project } = loadFixture('vanilla.pbxproj');
    addNotificationServiceExtensionToXcodeProject(project, baseOptions);

    const objects = project.hash.project.objects;
    assertBucketIsHomogeneous(objects.PBXTargetDependency, 'PBXTargetDependency');
    assertBucketIsHomogeneous(objects.PBXContainerItemProxy, 'PBXContainerItemProxy');
  });
``

File: __tests__/scenarios/ios/notificationServiceXcodeProject.test.ts [hunk 2]
```diff
@@ -149,6 +149,49 @@ describe('ios scenarios — addNotificationServiceExtensionToXcodeProject (vanil
       );
     }
   });
+
```
mahmoud-elmorabea, line 152: How about we also add a snapshot test for this? Helps catch any changes in the future. It will also be stable since its using a fixture.

mahmoud-elmorabea (commented):
Thanks for the fix! 

I think we should try to also upgrade our test app to Expo 55 and add `@bacons/xcode` as dep so we can catch these types of errors in the future.

__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
mrehan27: Updated

mrehan27:
@mahmoud-elmorabea Yeah we can update the test app to Expo 55 to prevent this going forward. But it shouldn't block this fix from going live since it's already been verified in a clean Expo 55 app. We can handle the test app update in a separate ticket, I'll create one for it.

__tests__/scenarios/ios/notificationServiceXcodeProject.test.ts
mrehan27: Good idea. But this should probably be added with Expo 55 update, so I'll include it in the same ticket.

mahmoud-elmorabea (approved):

PR #368, opened 2026-07-07
chore: add geofence build-time enablement (manual init)

mrehan27:
## Summary

First PR in the geofence stack. Adds **build-time enablement** of the Customer.io geofence native module via a new `geofence.enabled` plugin option. This wires the native module into the build and works with **manual initialization** — no auto-init (`config`) required. Auto-init config and the sample-app flow land in follow-up PRs.

Ticket: https://linear.app/customerio/issue/MBL-1787

> Targets `feature/geofence-on-device` (not `main`). This is part of a stacked series; the feature branch accumulates the geofence work until the native SDKs ship.

## What's included

**Plugin option**
- New `geofence?: { enabled?: boolean }` on `CustomerIOPluginOptions`. Off by default.
- Gated to **Expo SDK 53+** (Swift projects) — the plugin throws a clear error on older SDKs, since geofence relies on the Swift AppDelegate bootstrap.

**iOS**
- Adds the `geofence` subspec to the host-app Podfile. Geofence pulls in Location transitively, so the `location` subspec is omitted when geofence is enabled (avoids a redundant entry).
- Injects a `GeofenceModule.bootstrapForBackgroundDelivery(launchOptions:)` call at the top of `didFinishLaunchingWithOptions`, plus the `CioLocationGeofence` import. Both are wrapped in `#if canImport(CioLocationGeofence)` so they compile out when the subspec is absent. This is required so iOS cold-wake background geofence transitions are delivered even when the JS runtime isn't running. Injection is idempotent and handles the `internal import` variant (Expo 54).

**Android**
- Sets `customerio_geofence_enabled=true` in `gradle.properties`. The RN SDK enables the location module from this flag alone (geofence implies location), so no separate location flag is needed.

**Behavior notes**
- Geofence implies location on both platforms.
- Permissions and privacy keys (Info.plist `NSLocation*`, `UIBackgroundModes`; AndroidManifest location permissions) remain the **host app's responsibility** — the plugin injects none, matching the Location module and the SDK's "never request permission itself" contract.

## Tests
- `withGeofenceAppDelegate`: import placement, bootstrap-before-RN, `canImport` guards, `internal import` variant, idempotency.
- `withGeofenceGradleProperties`: add / update / disabled / omitted.
- `injectCIOPodfileCode`: full subspec matrix incl. geofence + push, geofence-implies-location.
- `withCIOIos`: geofence-only, geofence+location, geofence+push, disabled.
- `index`: Expo <53 throw, forwarding, disabled-no-throw.
- API report regenerated (adds `CustomerIOPluginGeofenceOptions` + the `geofence` option).

All green; `lint`, `tsc`, and the API-change check pass.

---

> [!NOTE]
> **Medium Risk**
> Changes native build wiring and modifies AppDelegate launch order for background geofence delivery; mistakes could affect cold-start behavior on iOS, though changes are gated to Expo 53+ and heavily tested.
> 
> **Overview**
> Adds a **`geofence?: { enabled?: boolean }`** plugin option (off by default) so Expo apps can compile in the Customer.io geofence native module without auto-init. **`geofence.enabled` requires Expo SDK 53+**; the entry plugin throws on older SDKs because geofence depends on Swift AppDelegate changes.
> 
> On **iOS**, when enabled, the Podfile gains the **`geofence`** subspec (combined with push subspecs as needed), and the redundant **`location`** subspec is skipped because geofence implies location. **`withGeofenceAppDelegate`** injects a **`#if canImport`-guarded** `CioLocationGeofence` import and **`GeofenceModule.bootstrapForBackgroundDelivery`** at the start of **`didFinishLaunchingWithOptions`**, before React Native starts, with independent idempotency for import vs bootstrap.
> 
> On **Android**, **`customerio_geofence_enabled=true`** is written to **`gradle.properties`** when geofence is on (same add/update pattern as location). Permissions and privacy strings are still the host app’s responsibility.
> 
> Public types and API extractor output are updated; tests cover Gradle properties, Podfile subspec matrix, iOS wiring, AppDelegate injection, and Expo version gating.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88ddf3b37dccc1dc397c490e7224e60bdfb8f2d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/plugin/src/ios/withGeofenceAppDelegate.ts
+++ b/plugin/src/ios/withGeofenceAppDelegate.ts
@@ -1,0 +1,105 @@
+import type { ConfigPlugin } from '@expo/config-plugins';
+import { withAppDelegate } from '@expo/config-plugins';
+
+import { logger } from '../utils/logger';
+
+// Per-step idempotency markers. The two injections are guarded independently: the import marker
+// is added by addGeofenceImport and the bootstrap marker by injectGeofenceBootstrap. Keying both
+// steps off a single symbol would let a partial application (import added, bootstrap injection
+// failed) look "done" on the next prebuild, permanently skipping the missing bootstrap.
+const GEOFENCE_IMPORT_MARKER = 'import CioLocationGeofence';
+const GEOFENCE_BOOTSTRAP_MARKER = 'GeofenceModule.bootstrapForBackgroundDelivery';
+
+const GEOFENCE_IMPORT_SNIPPET = `#if canImport(CioLocationGeofence)
+import CioLocationGeofence
+#endif`;
+
+const GEOFENCE_BOOTSTRAP_SNIPPET = `    #if canImport(CioLocationGeofence)
+    // iOS can cold-launch the app for a geofence transition without the JS runtime, so bootstrap
+    // background delivery here rather than relying on CustomerIO.initialize. Safe alongside normal init.
+    GeofenceModule.bootstrapForBackgroundDelivery(launchOptions: launchOptions)
+    #endif
+`;
+
+/**
+ * Adds the geofence module import after the last top-level import statement.
+ * The import is compiled out via `#if canImport` when the geofence subspec is absent.
+ */
+const addGeofenceImport = (contents: string): string => {
+  if (contents.includes(GEOFENCE_IMPORT_MARKER)) {
+    return contents;
+  }
+
+  const importRegex = /^(?:@_exported\s+)?(?:internal\s+)?import\s+.+$/gm;
+  let lastMatchEnd = -1;
+  let match: RegExpExecArray | null;
+  while ((match = importRegex.exec(contents)) !== null) {
+    lastMatchEnd = match.index + match[0].length;
+  }
+
+  if (lastMatchEnd < 0) {
+    logger.warn('Could not find an import statement in AppDelegate.swift; skipping geofence import');
+    return contents;
+  }
+
+  return (
+    contents.substring(0, lastMatchEnd) +
+    `\n\n${GEOFENCE_IMPORT_SNIPPET}` +
+    contents.substring(lastMatchEnd)
+  );
+};
+
+/**
+ * Injects the geofence background-delivery bootstrap at the top of didFinishLaunchingWithOptions,
+ * before React Native is started (the Expo template calls `factory.startReactNative(...)` later in
+ * this method). Cold-wake delivery must not depend on the JS runtime, so it has to run first —
+ * matching the SDK's reference AppDelegate integration.
+ */
+const injectGeofenceBootstrap = (contents: string): string => {
+  if (contents.includes(GEOFENCE_BOOTSTRAP_MARKER)) {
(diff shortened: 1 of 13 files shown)


plugin/src/ios/withGeofenceAppDelegate.ts
```diff
@@ -0,0 +1,105 @@
...
+import { withAppDelegate } from '@expo/config-plugins';
+
+import { logger } from '../utils/logger';
+
+// Per-step idempotency markers. The two injections are guarded independently: the import marker
+// is added by addGeofenceImport and the bootstrap marker by injectGeofenceBootstrap. Keying both
+// steps off a single symbol would let a partial application (import added, bootstrap injection
+// failed) look "done" on the next prebuild, permanently skipping the missing bootstrap.
+const GEOFENCE_IMPORT_MARKER = 'import CioLocationGeofence';
+const GEOFENCE_BOOTSTRAP_MARKER = 'GeofenceModule.bootstrapForBackgroundDelivery';
```
mahmoud-elmorabea, line 11: Should we just use our own `CioSdkAppDelegateHandler` for this instead of injecting this directly to AppDelegate?

mahmoud-elmorabea: I just realized that the timing of the call is different but I suggest we add a `CioGeofenceAppDelegateHandler.swift` that we can reduce the edits to AppDelegate to single call that internally does whatever we want that we can extend in the future

mahmoud-elmorabea (approved):

plugin/src/ios/withGeofenceAppDelegate.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const GEOFENCE_IMPORT_SNIPPET = `#if canImport(CioLocationGeofence)
+import CioLocationGeofence
+#endif`;
+
+const GEOFENCE_BOOTSTRAP_SNIPPET = `    #if canImport(CioLocationGeofence)
+    // iOS can cold-launch the app for a geofence transition without the JS runtime, so bootstrap
+    // background delivery here rather than relying on CustomerIO.initialize. Safe alongside normal init.
+    GeofenceModule.bootstrapForBackgroundDelivery(launchOptions: launchOptions)
```
mahmoud-elmorabea, line 20: From this comment, it's not clear why our bootstrap call has to be before `factory.startReactNative`

plugin/src/ios/withGeofenceAppDelegate.ts
mrehan27: Good call. Noted, will include this in the follow up PR.

plugin/src/ios/withGeofenceAppDelegate.ts
mrehan27: Will address this too in the follow up PR.

PR #372, opened 2026-07-09
chore: fix register CustomerIO Xcode group by key

mrehan27:
## What

`getOrCreateCustomerIOGroup` now returns the CustomerIO **group key** in both branches. Previously the create branch returned a key (correct) but the existing-group branch returned the group **object**.

## Why

`xcodeProject.addSourceFile(path, opt, group)` expects `group` to be a **key string**. When passed the object, the lookup falls through to `getPBXVariantGroupByKey`, which throws `Cannot read properties of undefined` on any `project.pbxproj` that has no `PBXVariantGroup` section — the case for the Expo SDK 55 template.

Today push is the only caller (creates the group once, reuses the key), so the bug is latent. It surfaces the moment a **second** caller finds the already-created group and gets the object back — which is exactly what the upcoming geofence AppDelegate handler does. This lands the fix first so that change stacks cleanly.

## Verification

- `expo prebuild` on Expo SDK 55 (test-app, RN 0.83.6): ✔ finishes; push handler registered (build file + file ref + CustomerIO group + Sources phase).
- lint, api-extractor (no public API change), and unit tests green (baseline pre-existing fcm/postInstall failures only).

Ticket: https://linear.app/customerio/issue/MBL-1787

---

> [!NOTE]
> **Medium Risk**
> Touches Expo config-plugin iOS `pbxproj` mutation during prebuild; incorrect group handling can break source registration, but the change is a narrow bugfix with targeted tests.
> 
> **Overview**
> Fixes iOS Expo prebuild so the CustomerIO Xcode group is always passed to `addSourceFile` as a **group key string**, not a group object.
> 
> `getOrCreateCustomerIOGroup` now resolves an existing group via `findPBXGroupKey` and returns that key; the create path still returns the key from `pbxCreateGroup`. `copyFileToXcode` and `withCIOIosSwift` typings are updated to treat the group as `string`, matching what `xcodeProject.addSourceFile` expects.
> 
> Adds **`__tests__/utils/xcode.test.ts`** with fixture-based regression tests for the existing-group branch and for `copyFileToXcode` passing a string into `addSourceFile`—covering the Expo SDK 55 case where missing `PBXVariantGroup` caused a crash when the object was passed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5255d4616f2a63a98cb69a4c6fa0041c150f37b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/plugin/src/utils/xcode.ts
+++ b/plugin/src/utils/xcode.ts
@@ -7,20 +7,21 @@ import { logger } from './logger';
  * Gets an existing CustomerIO group or creates a new one in the Xcode project
  * @param xcodeProject The Xcode project instance
  * @param projectName The iOS project name
- * @returns The CustomerIO group reference
+ * @returns The CustomerIO group key
  */
 export function getOrCreateCustomerIOGroup(
   xcodeProject: XcodeProject,
   projectName: string,
-): XcodeProject['pbxCreateGroup'] {
-  // Check if CustomerIO group already exists
-  let customerIOGroup = xcodeProject.pbxGroupByName('CustomerIO');
-  if (customerIOGroup) {
-    return customerIOGroup;
+): string {
+  // Return the group key (not the group object): addSourceFile expects a key, and passing the
+  // object throws on pbxproj files that have no PBXVariantGroup section (e.g. Expo SDK 55).
+  const existingKey = xcodeProject.findPBXGroupKey({ name: 'CustomerIO' });
+  if (existingKey) {
+    return existingKey;
   }
 
   // Create new CustomerIO group and add it to the project
-  customerIOGroup = xcodeProject.pbxCreateGroup('CustomerIO');
+  const customerIOGroup = xcodeProject.pbxCreateGroup('CustomerIO');
   const projectGroupKey = xcodeProject.findPBXGroupKey({ name: projectName });
   xcodeProject.addToPbxGroup(customerIOGroup, projectGroupKey);
   return customerIOGroup;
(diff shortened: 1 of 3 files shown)


plugin/src/utils/xcode.ts
```diff
@@ -7,20 +7,21 @@ import { logger } from './logger';
...
   projectName: string,
-): XcodeProject['pbxCreateGroup'] {
-  // Check if CustomerIO group already exists
-  let customerIOGroup = xcodeProject.pbxGroupByName('CustomerIO');
-  if (customerIOGroup) {
-    return customerIOGroup;
+): string {
+  // Return the group key (not the group object): addSourceFile expects a key, and passing the
+  // object throws on pbxproj files that have no PBXVariantGroup section (e.g. Expo SDK 55).
+  const existingKey = xcodeProject.findPBXGroupKey({ name: 'CustomerIO' });
```
Shahroz16, line 18: Could we add a focused regression test for this existing-group branch? The current `withCIOIosSwift` test mocks `getOrCreateCustomerIOGroup`, so it would not catch the object-versus-key bug if this regressed. Even a small fake project asserting that an existing CustomerIO group returns its key and that key is passed to `addSourceFile` would pin the behavior.

Shahroz16 (approved):
None blocking suggestion: maybe we need, a small test-coverage comment for the exact existing-group regression. The implementation itself matches the xcode library key contract.

plugin/src/utils/xcode.ts
mrehan27: Added

PR #376, opened 2026-07-17
chore: align sample location screen with RN sample

mrehan27:
Aligns the Expo test-app location screen with the customerio-reactnative sample ([mbl-2148-align-location-mode](https://github.com/customerio/customerio-reactnative/tree/mbl-2148-align-location-mode)), so both demos behave the same for geofence.

## Changes (test-app/screens/Location.js)
- **Order matches RN** — "Background Location (Geofence)" is now the first card (was last).
- **One-tap background escalation** — tapping "Grant location access" requests foreground, then immediately shows the background rationale and requests "Always" / "Allow all the time". Previously the flow stopped after foreground and required a second tap, so background was never prompted and users had to enable it manually in Settings.
- **State-aware button** — `Grant location access` → `Upgrade to 'Always'` / `Allow all the time` → `✓ granted`.
- **Sequence guard** on the permission refresh so a stale read can't overwrite a newer one (also resolves the earlier Bugbot stale-read note).
- **Rationale copy** — dropped the redundant Open Settings button (Continue already reaches the system prompt / Settings).

Sample-app only; no plugin code changes. On Android 11+ the OS forbids an in-app background-location prompt, so Continue routes to Settings there (expected).

Ticket: https://linear.app/customerio/issue/MBL-2148

---

> [!NOTE]
> **Low Risk**
> Sample-app screen only; no SDK or plugin changes—behavior risk is limited to the Expo demo’s permission UX.
> 
> **Overview**
> Aligns **`test-app/screens/Location.js`** with the customerio-reactnative sample so both demos exercise geofence the same way.
> 
> **Background location / geofence** is now the **first** section (removed the old “OPTION 5” placement and OR dividers). Permission handling is renamed and simplified (`locationStatus` drops `backgroundDenied`; denied is unified). A **`grantRank`** helper triggers `CustomerIO.geofence.refreshFromCurrentLocation()` only when access **increases** (e.g. after changing permissions in Settings), not on every resume.
> 
> **`refreshLocationStatus`** replaces the old geofence status reader and adds a **sequence guard** so overlapping mount/AppState/dialog refreshes cannot apply stale results. **`handleBackgroundLocationTap`** implements a state-aware flow: one tap from undetermined requests foreground then immediately shows the background rationale and continues to “Always” / “Allow all the time”; foreground-only upgrades via rationale; denied opens Settings. Success paths call **`onBackgroundGranted`** (refresh + alert). SDK and device location handlers now **`await refreshLocationStatus()`** after foreground requests.
> 
> UI copy and labels are platform-specific; section headings are flattened (SDK, presets, current location, manual entry). **`Button`** is imported from `react-native` with **`Platform`**; layout uses **`gap`** instead of OR rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b34f12a6366e80cab5c979a2c694119440eec2ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test-app/screens/Location.js
+++ b/test-app/screens/Location.js
@@ -24,8 +24,10 @@ const PRESETS = [
   { label: '0, 0', lat: 0, lng: 0 },
 ];
 
-const isGrantedStatus = (status) =>
-  status === 'foregroundOnly' || status === 'backgroundGranted';
+// Higher = more access. Used to fetch only when the permission level increases
+// (e.g. gaining "Always" in Settings), never on a downgrade between states.
+const grantRank = (status) =>
+  status === 'backgroundGranted' ? 2 : status === 'foregroundOnly' ? 1 : 0;
 
 function showLocationPermissionAlert() {
   Alert.alert(
@@ -103,10 +105,10 @@ export default function LocationScreen() {
       }
       const previous = locationStatusRef.current;
       const status = await refreshLocationStatus();
-      // Fetch only when gaining access from a non-granted state — not on a downgrade
-      // between granted states. The SDK's auto-fetch hook runs once per process, so a
-      // grant made in Settings needs an explicit request to register geofences.
-      if (status && isGrantedStatus(status) && !isGrantedStatus(previous)) {
+      // Fetch whenever the permission level increased (a grant made in Settings),
+      // including foregroundOnly → backgroundGranted; skip downgrades. The SDK's
+      // auto-fetch hook runs once per process, so Settings grants need an explicit fetch.
+      if (status && grantRank(status) > grantRank(previous)) {
         refreshGeofences();
       }
     });
--- a/test-app/screens/Location.js
+++ b/test-app/screens/Location.js
@@ -131,6 +131,11 @@ export default function LocationScreen() {
     try {
       const background = await Location.requestBackgroundPermissionsAsync();
       const next = await refreshLocationStatus();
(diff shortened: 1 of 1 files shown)


test-app/screens/Location.js
```diff
@@ -40,114 +46,183 @@ export default function LocationScreen() {
...
+      case 'denied':
+        Linking.openSettings();
+        return;
+      default: {
+        // notDetermined: request foreground first, then escalate to background.
         const foreground = await Location.requestForegroundPermissionsAsync();
-        if (foreground.status !== 'granted') {
-          setGeofenceStatus(foreground.canAskAgain ? 'notDetermined' : 'denied');
+        const next = await refreshLocationStatus();
+        if (next === null) {
```
Shahroz16: Could we keep this foreground-grant flow alive if its status refresh is superseded? If the foreground permission dialog produces an AppState active event after this handler starts its refresh, that newer refresh can make next null here. The winning AppState path refreshes geofences but does not call showBackgroundRationale, so the user needs a second tap to request Always. Would retrying the latest status here, or carrying a pending escalation into the winning refresh, help preserve the one-tap flow? The state remains correct, so I do not see this as blocking, but it seems worth keeping the sample behavior consistent.

Shahroz16 (approved):

PR #377, opened 2026-07-23
feat(live-notifications): configure Live Notifications through the plugin

mahmoud-elmorabea:
Adds a `liveNotifications` config surface and generates the native setup for it: activity-type registration in the auto-init initializer on **both** platforms, the iOS widget extension with branding compiled in, and the opened metric in the generated `openURL`.

Previously the generated initializer applied no Live Notifications config, so an auto-init app registered no types and never got a push-to-start token. Replaces the earlier `ios.liveActivity` option.

Depends on customerio-ios#1178.

---

> [!NOTE]
> **High Risk**
> Large native prebuild changes (Podfile, Xcode targets, AppDelegate/deep links, generated initializers) affect build and runtime routing; unreleased git-sourced iOS pods add release and dependency risk.
> 
> **Overview**
> Introduces **`liveNotifications`** as the plugin config surface (replacing the earlier **`ios.liveActivity`** option) and wires **Customer.io Live Notifications** into Expo prebuild on **iOS and Android**.
> 
> **iOS** turns the feature on via **`liveactivities`** Pod subspecs, **`NSSupportsLiveActivities`**, a generated **`CIOLiveActivityWidget`** extension (built-in templates + optional **`customWidget`** / branding), and **Live Activity tap** handling through **`application(_:open:options:)`** (including upgrades when push handler code already exists). **Android** patches the auto-init initializer for enabled types/branding, copies optional **`customRenderer`** sources, installs branding drawables, and registers the renderer from **`MainApplication`** when the SDK is initialized from JavaScript.
> 
> **Auto-initialization** now registers activity types in generated native initializers (fixing missing push-to-start registration). **Temporary** Podfile entries pull unreleased **customerio-ios** Live Activities pods from a feature branch until they ship on CocoaPods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b26c316034d0877b5d9175bbd377ed007d6fb5f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


plugin/src/helpers/utils/patchLiveNotificationCode.ts
```diff
@@ -0,0 +1,461 @@
...
+}
+
+function androidBrandingCall(
+  branding: LiveNotificationBranding | undefined
+): string | null {
+  if (!branding) return null;
+
+  const accent = branding.accentColorHex
+    ? `Color.parseColor("${branding.accentColorHex}")`
+    : 'Color.TRANSPARENT';
```
Shahroz16, line 466: `accentColorHex` is optional here too, so could we preserve the absence of a color instead of generating `Color.TRANSPARENT` for a logo or company-only branding block? Android treats that zero as its default color, so this does not make the UI invisible, but it bypasses an app's configured `default_notification_color`. This is non-blocking visual consistency feedback; ideally the native branding contract stays nullable so the app-level fallback remains effective.

plugin/src/types/cio-types.ts
```diff
@@ -141,6 +146,186 @@ export type CustomerIOPluginOptions = {
...
+   * Lives here rather than under `config.liveNotifications` because it is purely build-time —
+   * a source file to compile into the generated widget extension and a struct to instantiate in
+   * its `WidgetBundle`. It has no runtime meaning, and keeping it here is what makes it usable
+   * whether you initialize the SDK automatically or from JavaScript.
+   *
+   * The activity's identifier is separate and belongs to SDK config: set
+   * `config.liveNotifications.customType` for automatic initialization, or pass it to
+   * `CustomerIO.initialize` when you initialize from JavaScript.
+   */
+  customWidget?: LiveNotificationCustomWidget;
```
Shahroz16, line 197: Could we document the one thing the plugin cannot inject into a customer's custom SwiftUI: both the lock-screen root and Dynamic Island need `.cioWidgetUrl(context.state.cioMetadata)`. Without it, the generated AppDelegate URL handler is never invoked for custom activities, so every server-pushed custom tap loses opened attribution and deep-link routing. The test app in #380 currently demonstrates this gap.

Shahroz16 (approved):
I confirmed the existing no-push URL thread is real but narrower than the normal push-enabled flow: it matters for backend-driven Live Activities when an app registers a token through another provider and passes it to Customer.io. I would fix it, but I would not describe every no-push setup as broken. The extra push-pod thread is follow-up-level dependency overhead. I found two other direct API/config issues below; the Android tint one is non-blocking. One release prerequisite also remains outside this diff: the package still peers on React Native SDK 6.5.2, which does not expose this Live Activities API, so that needs to move to the release containing RN #617 before this plugin ships.

PR #380, opened 2026-07-27
feat(test-app): Live Notifications screen driven by auto-init config

mahmoud-elmorabea:
Exercises the plugin end to end: types and branding come from `config.liveNotifications`, so the screen needs no JavaScript initialization of its own.

Stacked on #377.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the test app, sample native renderers, and CI signing metadata; no production SDK or plugin logic in this diff.
> 
> **Overview**
> Adds an end-to-end **Live Activities** flow in the Expo test app so the plugin’s `liveNotifications` auto-init (types, branding, custom widget/renderer) can be exercised without extra JS setup.
> 
> **`app.json`** now registers built-in segment/countdown types plus a custom rideshare type, with branding and pointers to `RideshareLiveActivity.swift` and `RideshareLiveNotification.kt`. A new **`LiveActivities`** screen drives `CustomerIO.liveActivities` start/update/end for all three templates; navigation is wired from the dashboard. **Fastlane** includes the generated `CIOLiveActivityWidget` extension for ad-hoc signing. **`customerio-reactnative`** is pinned to the `live-activities-sdk` branch for this demo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 340eb0da42459c8ec3814e5c9782041513213a39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test-app/app.json
+++ b/test-app/app.json
@@ -100,7 +100,8 @@
                 "accentColorHex": "#00A0DF",
                 "backgroundColorHex": "#101010",
                 "textColorHex": "#FFFFFF"
-              }
+              },
+              "customType": "com.customerio.testbed.rideshare"
             }
           },
           "android": {
@@ -133,6 +134,12 @@
           },
           "location": {
             "enabled": true
+          },
+          "liveNotifications": {
+            "customWidget": {
+              "sourceFile": "./ios-widgets/RideshareLiveActivity.swift",
+              "structName": "RideshareLiveActivity"
+            }
           }
         }
       ],
--- a/test-app/app.json
+++ b/test-app/app.json
@@ -139,6 +139,10 @@
             "customWidget": {
               "sourceFile": "./ios-widgets/RideshareLiveActivity.swift",
               "structName": "RideshareLiveActivity"
+            },
+            "customRenderer": {
+              "sourceFile": "./android-renderers/RideshareLiveNotification.kt",
+              "className": "RideshareLiveNotificationCallback"
             }
           }
         }
--- a/test-app/app.json
+++ b/test-app/app.json
@@ -89,6 +89,18 @@
             "logLevel": "debug",
             "location": {
               "trackingMode": "MANUAL"
+            },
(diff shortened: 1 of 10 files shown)


test-app/app.json
```diff
@@ -89,6 +89,13 @@
             "logLevel": "debug",
             "location": {
               "trackingMode": "MANUAL"
+            },
+            "liveNotifications": {
+              "types": [
+                "io.customer.livenotifications.segments",
+                "io.customer.livenotifications.countdowntimer"
+              ],
+              "customType": "com.customerio.testbed.rideshare"
```
Shahroz16, line 98: Could we either add an Android `CustomerIOLiveNotificationsCallback` for this type or make the custom demo iOS-only? The generated Android initializer allowlists this id, so `start()` returns an id, shows Success and can emit a backend start, but Android has no built-in renderer for custom types and this repo never registers a callback. Every Android custom start therefore posts nothing.

test-app/ios-widgets/RideshareLiveActivity.swift
```diff
@@ -0,0 +1,51 @@
...
+                    Text(context.state.data["status"] ?? "")
+                }
+            } compactLeading: {
+                Image(systemName: "car.fill")
+            } compactTrailing: {
+                Text("\(context.state.data["etaMinutes"] ?? "")m")
+            } minimal: {
+                Image(systemName: "car.fill")
+            }
+        }
```
Shahroz16, line 54: Could we apply `.cioWidgetUrl(context.state.cioMetadata)` to both the lock-screen view and the `DynamicIsland` here? Without it the custom widget never emits the Customer.io tap URL, so the generated AppDelegate handler cannot report an opened metric or route the customer deep-link for Rideshare.

Shahroz16 (approved):

PR #385, opened 2026-08-10
ci: remove the dead npm unpublish workflow

mahmoud-elmorabea:
Removes `unpublish-npm-version.yml` and its only caller-script, both of which depend on an `NPM_TOKEN` that npm's token restrictions have already disarmed.

Fixes MBL-2249

[MBL-2249: Remove the dead npm token workflows](https://linear.app/customerio/issue/MBL-2249/remove-the-dead-npm-token-workflows)

### Why this is safe

- The workflow is `workflow_dispatch`-only and has **never been run** — zero runs in its entire history.
- It authenticates via the `NPM_TOKEN` secret. npm's phase-1 token restrictions (31 July 2026) removed *management* capabilities from these tokens, so `npm unpublish` through it is most likely already non-functional.
- Nothing else in the repo references the workflow or its script.

### Why the script goes too

`scripts/unpublish-npm-version.sh` had exactly one caller — the deleted workflow — and writes `//registry.npmjs.org/:_authToken=${NPM_TOKEN}` into an `.npmrc`. `git grep NPM_TOKEN` returns nothing in this repo after this change, which is the point: the milestone is that no release depends on a restricted token, ahead of npm's phase-2 restrictions (~January 2027).

### What we give up

The ability to unpublish a version via CI — never used once, and probably already dead. If it is ever needed it becomes a deliberate, human, 2FA-gated action, which is the right shape for an irreversible registry operation and a better one than a dispatchable workflow holding a long-lived token.

Publishing is unaffected: it runs in `deploy-sdk.yml` and does not use `NPM_TOKEN`.

### Follow-up (not in this PR)

The `NPM_TOKEN` repo secret can be deleted and the underlying token revoked once this has merged.

---

> [!NOTE]
> **Low Risk**
> Deletion-only CI cleanup with no runtime or publish-path changes; unpublish was workflow_dispatch-only and unused.
> 
> **Overview**
> Removes the manual **`unpublish-npm-version`** GitHub Actions workflow and **`scripts/unpublish-npm-version.sh`**, which were the only in-repo paths that used **`NPM_TOKEN`** to run **`npm unpublish`** on **`customerio-expo-plugin`**.
> 
> This drops CI-based version unpublishing (never used) and clears the repo of references to a token whose npm management capabilities are already restricted. **Publishing** via **`deploy-sdk.yml`** is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ade148375f82a97f3643609a9bb48181ae584df6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jcpsimmons (approved):
Checked the follow-up claim myself: deploy-sdk.yml publishes via OIDC (id-token: write, npm pinned to 11.x for exactly that), so nothing left in the repo touches NPM_TOKEN after this. Safe to drop the secret and revoke the token once it merges.

mrehan27 (approved):

PR #390, opened 2026-08-14
ci: hold the expo compatibility app to the stable expo release

mahmoud-elmorabea:
The Expo compatibility job could generate a test app running a pre-stable `expo` release, turning every open PR red for reasons outside this repo. It now holds the app to the release on the `latest` dist-tag, resolved fresh on every run.

Not a committed pin — nothing is stored here and nothing needs bumping. The day a release is promoted to `latest`, the next run tests against it, so a genuinely incompatible release still fails CI immediately. Only releases that haven't reached `latest` are excluded, and those aren't what customers install.

Two independent places decide the version, so both are handled:

1. **Template choice** — `sdk-<major>` tracks `next`, so it can point at a template pinning an `expo` that isn't out yet. The curated tag is still preferred; the script walks back to the newest usable template only when the tag is unusable.
2. **The version itself** — choosing the template is *not* sufficient. npm resolves a `~` range to the highest published version regardless of dist-tag, so `~57.0.12` still installs 57.0.13 when that release exists on `next` only. The exact stable version is written into the app before install, and re-asserted in `setup-test-app` after `npx expo install --fix`, which is an independent second writer and otherwise overwrites it.

An unsatisfiable major is reported as an upstream registry state rather than a plugin regression; a bad `--expo-template` is reported as our error.

Applies to all three consumers of these scripts: the PR smoke matrix, the release matrix, and the nightly iOS toolchain workflow.

**Verified end to end** on a real generated app with the divergence forced: `--fix` attempts the bump, the restore wins, the final installed version is the stable one.

**Known limitation** (documented in the README): only `expo` is held to stable. The template's other ~20 `expo-*` pins still float, so a broken pre-stable release of one of those can still fail the install.

### DIFF
--- a/__tests__/scripts/expo-template.test.js
+++ b/__tests__/scripts/expo-template.test.js
@@ -1,0 +1,296 @@
+const {
+  templatePackageName,
+  parseTemplateCandidates,
+  selectTemplateVersion,
+  chooseStableExpoVersion,
+  describeTemplateResolutionFailure,
+} = require('../../scripts/utils/expo-template');
+
+// Real `npm view 'expo-template-default@>=57.0.0 <58.0.0' version dependencies.expo --json`
+// output, captured 2026-08-14. Each template pins the `expo` release from two
+// patches earlier, which is what makes the newest template unusable while
+// `expo@latest` catches up.
+const TEMPLATES_57 = [
+  { version: '57.0.0', expoRange: '~57.0.0-preview.0' },
+  { version: '57.0.1', expoRange: '~57.0.0-preview.1' },
+  { version: '57.0.2', expoRange: '~57.0.0' },
+  { version: '57.0.3', expoRange: '~57.0.1' },
+  { version: '57.0.4', expoRange: '~57.0.2' },
+  { version: '57.0.5', expoRange: '~57.0.3' },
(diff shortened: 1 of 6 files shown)


__tests__/scripts/expo-template.test.js
```diff
@@ -0,0 +1,372 @@
...
+
+  it('uses the sdk-<major> dist-tag for an older major', () => {
+    const src = sources({ latest: '57.0.13', sdkTag: '54.0.36' });
+
+    expect(chooseStableExpoVersion(54, src)).toBe('54.0.36');
+    expect(src.calls).toEqual(['latest', 'sdkTag:54']);
+  });
+
+  it('falls back to the newest stable release for a major with no sdk tag', () => {
+    // `expo` prunes old channel tags: sdk-52 is the oldest that still exists,
```
mrehan27, line 210: We don't really need to handle or validate versions that we no longer support or include in our compatibility testing. I think we can drop those cases and simplify the code.

mrehan27 (approved):
The problem seems fairly limited with a small window for it to occur, and it only affects CI, not customers. I think it would be better if we could simplify the handling here, since the current approach adds some maintenance cost for us. However, if simplifying it requires significant rework, I'm okay with keeping it as is for now.

PR #391, opened 2026-08-17
ci: verify Xcode 27 generated app launches

Shahroz16:
## Summary

- Run the generated Expo APN and FCM apps with Node 24 on the floating Xcode 27 preview runner.
- Resolve the actual generated workspace and current simulator product, reject stale DerivedData products, then install, launch, and require short process survival.
- Use the merged `mobile-ci-tools` launch action at `@main` as the single owner of launch classification, summary, and bounded simulator logs.
- Keep repository Actions history as the monitoring source of truth; no cross-repository watchdog or duplicated reporting layer is included.

## Release impact

CI-only. This PR changes no plugin runtime source, package metadata, or release artifact.

## Current expected result

The Xcode 27 APN and FCM launch jobs are expected to remain red until the Expo UIScene production adapter lands. That red result is the product gap this smoke test is meant to expose, not a CI defect to suppress.

## Validation

- focused workflow Jest suite: 2/2
- generated-product resolver behavior suite: 17/17
- Prettier, actionlint, and `git diff --check`
- all ordinary exact-head checks passed; the Xcode 27 APN and FCM launch checks intentionally fail with the verified UIScene-required signature

## Boundary

This proves generated-project compilation, simulator installation, launch, executable identity, and short process survival. It does not prove lifecycle callback delivery, real APN/FCM delivery, signing, physical-device behavior, or App Store acceptance.

---

> [!NOTE]
> **Medium Risk**
> Changes are confined to preview CI and helper scripts but depend on an external launch action and hosted Xcode 27 runners; misconfiguration could block nightly signal or leak build settings without the sanitizer.
> 
> **Overview**
> Extends the **Xcode 27 preview compatibility** workflow from compile-only checks to **simulator install, launch, and a short process-survival window** (`LAUNCH_SURVIVAL_SECONDS`) for APN and FCM matrix jobs, using the shared `launch-simulator-app` action from `mobile-ci-tools`.
> 
> After the native build, a new **`resolve_simulator_app.py`** step discovers the single generated `.xcworkspace`, runs `xcodebuild -showBuildSettings`, and resolves the **current Release simulator `.app`** (rejecting stale DerivedData via a build-start epoch). Private build settings are stripped before upload; launch and sanitized diagnostics are published as **artifacts**. The job moves to **Node 24**, uses **`--clean`** when creating test apps, and PR triggers now include the resolver script.
> 
> **`__tests__/workflows/`** adds Jest coverage of the workflow shape plus the Python resolver unittest suite; **`TESTING.md`** and **`IOS-TOOLCHAIN-COMPATIBILITY.md`** document that **launch failures are expected** until Expo UIScene host-lifecycle support lands—that red state is the signal, not noise to suppress.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c949785708dbb99d43a12d64cf6503153897c43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Shahroz16:
Hosted Xcode 27 result: the new gate reproduced the real product failure for both APN and FCM. Each generated app completed prebuild/configuration/Xcode compilation, installed, returned a launch PID, and then exited after 1 second. Simulator logs report: `Application failed to launch: UIScene life cycle is required for apps built with this SDK.` Run: https://github.com/customerio/customerio-expo-plugin/actions/runs/32054992727\n\nThis confirms the CI correction is working and that the PR must remain draft/red until the Expo-generated host adopts the required scene lifecycle. We should not weaken or mark the launch step as an expected pass. The shared-action dependency also remains: mobile-ci-tools #16 must merge, then this reference changes to `@main`.

.github/IOS-TOOLCHAIN-COMPATIBILITY.md
```diff
@@ -1,19 +1,26 @@
...
+workflow records the hosted image, macOS, architecture, exact Xcode build, SDK versions, and
+installed runtimes through shared `mobile-ci-tools` actions. It verifies toolchain families rather
+than copying an exact beta-image pin into this repository. A hosted preview label can become
+unavailable before a job starts, and job timeouts do not cover queue time; neither state is a
+compatibility pass. If an expected result or watchdog alert is absent, check this repository's
+Actions history and the pending native watchdog PR before treating the nightly as healthy.
 
-When Xcode 27 becomes stable, remove this temporary preview workflow and add Xcode 27 to the normal required compatibility and release paths. Exact beta-image validation belongs in a temporary, explicitly test-only PR. A compatibility pass proves generated-project compilation for the named Expo/provider combination. It does not prove app launch, lifecycle callback forwarding, physical-device push delivery, signing, export, or App Store acceptance.
+The current generated APN and FCM apps are expected to report `generated-app-launch-failed` or
```
mahmoud-elmorabea: **This is the blocker.** Pre-declaring both classifications as expected makes every runtime regression indistinguishable from the known one.

`did-not-survive` → `generated-app-exited-early` fires for *any* process that dies inside the 10s window. Off the top of my head: a missing or corrupt embedded JS bundle, an `--add-default-config` CIO key that traps in SDK init, a static-frameworks Podfile change that only fails at dyld time, an unrelated Expo module crash. All of them land on "the classification the policy says to expect until Expo UIScene support lands," and the doc explicitly tells the reader not to treat it as a defect to chase.

So a real plugin regression that crashes the generated app on launch ships, and the nightly's red state provides cover for it. That's a worse position than the compile-only job this replaces, because it *looks* like runtime coverage.

What's needed is a discriminator. Grep the collected launch log for the UIScene rejection signature — `UIScene life cycle is required` or `_UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption` — and only classify as the known blocker when it's present; anything else gets its own classification and is genuinely actionable. customerio-flutter#395 does exactly this for its legacy negative control (`grep -Eq 'UIScene life cycle is required|_UIApplicationEvaluate…'`), so the pattern already exists in the org.

Smaller note on this same file: it says to check "the pending native watchdog PR," which is accurate — customerio-flutter#395's docs link `xcode27-nightly-watchdog.yml` as a live file and it 404s. Worth aligning the two.

scripts/compatibility/resolve_simulator_app.py
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env python3
```
mahmoud-elmorabea: Design question: this whole file, its 278-line test suite, the new `yaml` devDependency, the Jest workflow-contract test, and a second full `xcodebuild -showBuildSettings` run (30-90s on an Expo+CocoaPods workspace) exist to answer "where did the build put the .app."

The reason it's hard is that `scripts/compatibility/validate-plugin.js:120-124` builds with no `-derivedDataPath`:

```js
xcodebuild -workspace ${workspaceName}.xcworkspace \
  -scheme ${workspaceName} \
  -sdk iphonesimulator \
  -configuration Release build
```

Add `-derivedDataPath "$RUNNER_TEMP/dd"` and the product path is `$RUNNER_TEMP/dd/Build/Products/Release-iphonesimulator/<Scheme>.app` — deterministic, no settings dump, no resolver. It also makes the stale-product epoch guard below moot, since a per-job DerivedData directory cannot contain a previous run's product, and it removes the `Release`/`CONFIGURATION` filtering that has to mirror validate-plugin.js by hand.

I'd rather fix the shared build invocation than layer a special-cased resolver on top of it. If there's a reason the build can't take that flag, worth saying so in the PR body, because right now the resolver reads as accidental complexity.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -107,8 +170,126 @@ jobs:
...
+            did-not-survive)
+              classification=generated-app-exited-early
+              next_step='inspect the launch log for the UIKit runtime issue or early process exit'
+              ;;
+            invalid-input|invalid-app|sdk-mismatch)
+              classification=launch-contract-invalid
+              next_step='inspect the generated product metadata and smoke-test inputs'
+              ;;
+            *)
+              echo "::warning::unrecognized launch failure reason: $FAILURE_REASON"
```
mahmoud-elmorabea: `unexpected-error` is the one documented `failure-reason` value not covered here, so it falls into `*)` and gets annotated as unrecognized.

The action's vocabulary (`action.yml:44-46`, `mobile-ci-tools/README.md:27-31`) is: `invalid-input`, `invalid-app`, `sdk-mismatch`, `runtime-unavailable`, `runtime-selection-failed`, `simulator-boot-failed`, `install-failed`, `launch-failed`, `did-not-survive`, `unexpected-error`, `none`. You cover everything except the last non-`none` one, and `launch.sh` reaches it from three `fail unexpected-error` sites plus its global ERR trap — including paths that run after install and launch.

Concretely: a nightly where the runner image is missing something `launch.sh` needs prints `::warning::unrecognized launch failure reason: unexpected-error` (false — it's documented), classifies as the generic launch failure, and hands the maintainer next-step text about the shared action reference for what is actually a runner-image problem.

Same gap in customerio-ios#1216.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -107,8 +170,126 @@ jobs:
...
             echo "**Toolchain:** Xcode ${XCODE_VERSION:-unknown} build ${XCODE_BUILD:-unknown}, iphonesimulator ${IOS_SIMULATOR_SDK:-unknown}"
-            echo '**Scope:** generated-project compatibility build; no launch, callback, push-delivery, signing, or App Store claim'
+            echo "**Scope:** generated-project compile, simulator install, launch, and ${LAUNCH_SURVIVAL_SECONDS}-second process survival; no callback, push-delivery, signing, or App Store claim"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Record launch failure
+        if: failure() && steps.launch.outcome == 'failure'
+        shell: bash
+        env:
+          FAILURE_REASON: ${{ steps.launch.outputs.failure-reason }}
```
mahmoud-elmorabea: The whole classification design reads `steps.launch.outputs.failure-reason` off a composite action whose inner step exited non-zero, and the action's own README says that isn't guaranteed.

`mobile-ci-tools/README.md:33-35`:

> A failed composite action is not required to propagate its mapped outputs, so diagnostics upload must not depend on the `log-path` output. If the action uses `continue-on-error: true`, gate that upload on `steps.<id>.outcome == 'failure'` because the job-level `failure()` predicate remains false.

`failure-reason` is mapped off the same inner step as `log-path`, so the warning covers it. If the runner doesn't publish it, then for *every* real launch failure `FAILURE_REASON` is empty, the `''` arm fires, and every red nightly reports `launch-smoke-inconclusive` — the exact opposite of this PR's purpose.

I don't think this is settled anywhere: `mobile-ci-tools`' own `ios-actions-test.yml` asserts `failure-reason` only on the success path, and for its deliberate-crash fixture it uses `continue-on-error: true` and checks `outcome` plus log greps, never the output. So the failure-path propagation is untested upstream too.

Worth confirming empirically on a dispatch run before building nine classification arms on top of it — and keeping `steps.launch.outcome` as the primary gate either way. Same issue on customerio-ios#1216, where it also causes the crash report to be skipped.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -13,11 +14,12 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
```
mahmoud-elmorabea: Adding `github.event_name` to the concurrency group means `schedule-refs/heads/main` and `workflow_dispatch-refs/heads/main` are now distinct groups, so a manual dispatch no longer cancels an in-flight nightly.

Dispatching while the 07:17 nightly is queued or running gives four concurrent 90-minute `xcode-27` jobs (2 providers x 2 runs) competing for the preview label. Our macOS capacity is already an org-wide cap, and this PR's own policy doc notes that a label can become unavailable before a job starts and that job timeouts don't cover queue time — so the dispatched run is now more likely to sit behind the nightly it was meant to replace.

If the goal was stopping PR runs from cancelling nightlies, `github.ref` alone already handles it (`pull_request` runs use `refs/pull/N/merge`). Same change in customerio-ios#1216.

scripts/compatibility/resolve_simulator_app.py
```diff
@@ -0,0 +1,207 @@
...
+        executable = path / executable_name
+        if not executable.is_file():
+            rejected.append(f"{path}: missing executable {executable_name}")
+            continue
+        try:
+            executable_modified_at = executable.stat().st_mtime
+        except OSError as error:
+            rejected.append(f"{path}: unreadable executable {executable_name} ({error})")
+            continue
+        if executable_modified_at < build_started_at:
```
mahmoud-elmorabea: The freshness comparison mixes a float against a truncated integer with strict `<`, leaving a sub-second window where a stale product reads as fresh.

`executable.stat().st_mtime` is a float; `build_started_at` comes from `date +%s` at `.github/workflows/ios-toolchain-compatibility.yml:95`, which truncates. So a stale `.app` executable with `st_mtime = 100.9` from a previous run, with the epoch captured at wall-clock 100.95 and recorded as `100`, satisfies `100.9 < 100 == False` and is accepted. The launch action then installs and launches the *previous* run's binary while the summary reports this run's toolchain — a false pass or a false failure attributed to the wrong build.

Easy fixes: capture with `date +%s.%N` (or `python3 -c 'import time;print(time.time())'`) and compare with a small negative margin. Or make it unnecessary with a per-job `-derivedDataPath`, per my note at the top of this file.

scripts/compatibility/resolve_simulator_app.py
```diff
@@ -0,0 +1,207 @@
...
+)
+
+
+def is_settings_payload(candidate: object) -> bool:
+    """Return whether a decoded value is a non-empty xcodebuild target list."""
+
+    return (
+        isinstance(candidate, list)
+        and bool(candidate)
+        and all(
```
mahmoud-elmorabea: `all(...)` here means one malformed entry discards the entire settings payload — including the diagnostic the failure message tells you to go read.

An Expo+CocoaPods workspace dumps ~40 targets. If any single one emits an entry without a `buildSettings` key (or with it null), `is_settings_payload` returns False, `load_settings` takes the parse-error branch, and the sanitized `<provider>-build-settings.json` artifact contains only `{parse_error, raw_bytes}` — no target list at all. Then `generated-app-resolution-failed` instructs the maintainer to inspect exactly that artifact.

The app target's settings were present and parseable, so the resolve could have succeeded. A per-entry filter — skip malformed entries, count them into `skipped`, keep going — both resolves and preserves the diagnostic. Given `skipped`/`rejected` are already threaded through the error message, this looks like a small change.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -87,16 +90,76 @@ jobs:
...
+        id: resolve-app
+        shell: bash
+        run: |
+          set -euo pipefail
+          settings_json="$RUNNER_TEMP/${{ matrix.ios-push-provider }}-build-settings.json"
+          private_settings_json="$RUNNER_TEMP/${{ matrix.ios-push-provider }}-build-settings-private.json"
+          trap 'rm -f "$private_settings_json"' EXIT
+          build_start_epoch="$RUNNER_TEMP/${{ matrix.ios-push-provider }}-build-start-epoch"
+          shopt -s nullglob
+          workspaces=("$APP_PATH"/ios/*.xcworkspace)
```
mahmoud-elmorabea: Workspace/scheme discovery is now implemented twice, in two languages, with different semantics.

`scripts/compatibility/validate-plugin.js:39-46` already does it: `readdirSync(ios).filter(endsWith('.xcworkspace'))[0]`, falling back to `app.json`'s `expo.name`, then to the literal `App`. This bash version globs and hard-fails unless exactly one workspace exists, with no fallback.

So if `expo prebuild` ever emits a second `.xcworkspace` (or a stale one survives), validate-plugin builds `workspaces[0]` and succeeds while this step exits 1 with `expected one generated workspace, found 2`. And if `ios/` has no workspace, validate-plugin builds `App.xcworkspace` while this reports a different error. Two lookups that must agree, kept in sync by hand, with nothing testing that they do.

Related trigger gap: `paths:` covers this workflow and `resolve_simulator_app.py`, but **not** `scripts/compatibility/validate-plugin.js`. So a PR changing the build invocation (adding `-derivedDataPath`, switching to `-destination`, changing the configuration) doesn't run this workflow, and the coupling breaks silently until the next nightly reports `generated-app-resolution-failed` — which points the reader at the resolve log rather than at the build script that changed.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -107,8 +170,126 @@ jobs:
...
+          {
+            echo '## Expo Xcode 27 preview compatibility'
+            echo
+            echo '**Classification:** toolchain-unavailable'
+            echo "**Provider:** ${{ matrix.ios-push-provider }}"
+            echo '**Next step:** inspect the hosted Xcode 27 image and shared toolchain report before attributing this to Expo'
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Record unclassified failure
+        if: failure() && steps.toolchain.outcome != 'failure' && steps.validate-plugin.outcome != 'failure' && steps.resolve-app.outcome != 'failure' && steps.launch.outcome != 'failure' && steps.launch.outcome != 'success'
```
mahmoud-elmorabea: Replacing `if: always() && !cancelled()` with five mutually-exclusive `failure()` guards opens a hole: a job that fails *after* the launch step succeeded produces no classification summary at all.

`Record unclassified failure` requires `steps.launch.outcome != 'success'`, and `Record compatibility result` requires `== 'success'` with an implicit `success()` gate. So if the launch passes and then `Record compatibility result` itself fails (a `$GITHUB_STEP_SUMMARY` write error), or the 90-minute `timeout-minutes` fires during a later step, the run is red with zero `## Expo Xcode 27 preview compatibility` output — whereas `always() && !cancelled()` always emitted one.

The simulator boot + install + launch + log collection this PR adds makes the timeout path materially more likely than it was before. Worth keeping one unconditional catch-all.

__tests__/workflows/ios-toolchain-compatibility.test.js
```diff
@@ -0,0 +1,101 @@
...
+      __dirname,
+      '../../scripts/compatibility/resolve_simulator_app.py'
+    ),
+    'utf8'
+  );
+  const definition = parse(workflow);
+  const steps = definition.jobs.preview.steps;
+  const step = (name) => steps.find((candidate) => candidate.name === name);
+
+  it('uses the repository Node version and launches the generated app', () => {
```
mahmoud-elmorabea: ~30 assertions in one `it`, most of them `toContain` against the workflow's raw source text — so cosmetic edits fail the suite, and the first failure masks the structural checks that actually encode contracts.

`expect(workflow).toContain('trap \'rm -f "$private_settings_json"\' EXIT')`, `toContain('xcodebuild_status="$?"')`, and `toContain('workspaces=("$APP_PATH"/ios/*.xcworkspace)')` pin exact shell text including local variable names and quoting. Renaming `private_settings_json`, reflowing the `xcodebuild` continuation lines, or switching to `local -i status` all fail while behavior is identical. And because they share one `it`, the first mismatch aborts before the assertions on `step('Record launch failure').if` and the artifact path list — so a real regression in a classification condition can hide behind a whitespace diff.

The parsed `definition` is already in hand. I'd let the structural assertions carry the contract, split them into separate `it` blocks, and drop the text greps.

Also, line 61 asserts the literal `.../launch-simulator-app/v1@main`. That actively pins the *unpinned* ref, so upstream vocabulary drift (see my `unexpected-error` note) can't be caught here — it's asserting the thing that makes drift possible.

scripts/compatibility/test_resolve_simulator_app.py
```diff
@@ -0,0 +1,278 @@
+import json
```
mahmoud-elmorabea: This placement contradicts the rule stated in `TESTING.md:97`, which this PR also edits:

> The compatibility scaffolding scripts live in `scripts/compatibility/` (see its `README.md`). They're for the matrix and for manual local repro — **not for unit tests**.

A 278-line unittest suite now lives there. And `scripts/compatibility/README.md` still documents only the original five scripts with no mention of `resolve_simulator_app.py` — so a maintainer following TESTING.md's pointer to that README won't learn the resolver exists, what its four required arguments are, or that a Python suite covers it. The new `__tests__/workflows/` row you added to TESTING.md also implies workflow-helper tests live under `__tests__/`, which this file doesn't.

Pick one: move it under `__tests__/`, or update both TESTING.md's rule and the compatibility README to describe the new arrangement.

TESTING.md
```diff
@@ -4,44 +4,45 @@ This guide explains how the test suite is layered and **where new tests should g
...
-| `validate-plugin-compatibility-matrix.yml` | merge to `main` (release) + manual | Full multi-SDK matrix — real `expo prebuild` + native build |
-| `reusable_build_sample_apps.yml` | release / sample-app distribution | Full Android + iOS app builds shipped to Firebase |
+| Workflow                                   | Trigger                                      | Tests                                                                                                                                                                                  |
+| ------------------------------------------ | -------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `test.yml`                                 | every PR + push to `main`/`beta`/`feature/*` | All four Jest projects (parallel jobs)                                                                                                                                                 |
```
mahmoud-elmorabea: Two things here.

**The stale claim is being re-asserted.** This row says `every PR + push to main/beta/feature/*`, but `.github/workflows/test.yml:3-6` declares only `pull_request:` and `push: branches: [main]`. So a contributor pushing to `feature/*` expects the four Jest projects to run and gets nothing until they open a PR. Since this diff rewrites the line, it's worth correcting rather than reformatting.

**The Prettier churn buries the real changes.** Of 56 changed lines in this file, only the `ios-toolchain-compatibility.yml` row, the new `__tests__/workflows/` row, one `root-tests` sentence, and a code-fence reflow carry information; the rest is column padding across four tables. `git blame` on those rows now points at a CI PR, and unrelated doc PRs will conflict on padding. Worth splitting into its own commit or leaving out.

mrehan27 (approved):

PR #395, opened 2026-08-23
test: validate Expo scene lifecycle canary (MBL-2234)

Shahroz16:
## Test only, do not merge

Expo 57 stable remains AppDelegate-only. This PR pins an exact Expo 58 canary that generates a scene host, then runs the Customer.io APN and FCM configurations through Xcode 27 generation, build, install, and launch checks.

The generated code establishes two separate results:

- Expo now starts React Native from `ExpoAppSceneDelegate`, while Customer.io's existing killed-state launch-options rewrite remains in AppDelegate. That rewrite cannot supply scene cold-start notification options.
- The canary still terminates during Xcode 27 launch in Expo's own `UIApplication.statusBarFrame` path.

This branch does not add a production workaround, replace Expo's root scene, or pin a canary for customers. After the hosted evidence is recorded, close this PR without merging. The production handoff remains MBL-2234 and physical notification evidence remains MBL-2233.

### DIFF
--- a/.github/workflows/ios-toolchain-compatibility.yml
+++ b/.github/workflows/ios-toolchain-compatibility.yml
@@ -79,9 +79,6 @@ jobs:
           npx create-expo-app "$APP_PATH" \
             --template "expo-template-default@$EXPO_CANARY_VERSION" \
             --no-install
-          npm pkg set \
-            --prefix "$APP_PATH" \
-            "dependencies.expo=$EXPO_CANARY_VERSION"
           npm install --prefix "$APP_PATH"
 
       - name: Setup Test App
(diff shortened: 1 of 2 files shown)


Shahroz16:
Validation complete on exact Expo `58.0.0-canary-20260812-27f94d4`: both APN and FCM variants generated, passed Customer.io plugin validation, and built on Xcode 27. Both then exited after 3 seconds during launch, with the final app log at Expo upstream use of `UIApplication.statusBarFrame`, which UIKit reports as a no-op on iOS 27.

Evidence: https://github.com/customerio/customerio-expo-plugin/actions/runs/32648679145

This confirms two separate gates: the current Expo scene canary cannot survive Xcode 27 launch, and Customer.io killed-state launch-option rewriting remains in AppDelegate while Expo now starts React Native from `ExpoAppSceneDelegate`. This PR is test-only evidence and should not merge.

Shahroz16:
Closing as planned after recording the test-only evidence. No canary pin or validation workflow is being merged.

PR #397, opened 2026-08-24
feat: add Android push notification icon and color options (MBL-2297)

### LINKED ISSUE #115: Notification icon / color defined in the expo project are not used
It's will be great to add 2 entries on the plugin configuration : "android.icon" and "android.color"

jcpsimmons:
## Summary

In managed workflow there was no way to customize the Android push notification icon or accent color, so notifications rendered with the default grey square (#115). The only workaround was maintaining the `AndroidManifest.xml` entries by hand, which does not survive prebuild.

This adds two options under `android.pushNotification` and injects the corresponding Firebase meta-data entries into `AndroidManifest.xml` at prebuild:

```json
[
  "customerio-expo-plugin",
  {
    "android": {
      "pushNotification": {
        "icon": "./assets/notification-icon.png",
        "color": "#1DA1F2"
      }
    }
  }
]
```

- `icon`: a path to a local image (copied into `res/drawable` as `cio_notification_icon`; should be white-on-transparent since Android renders status bar icons in a single color), or a reference to a drawable the app already defines (`@drawable/ic_stat_notification`). Written to the manifest as `com.google.firebase.messaging.default_notification_icon`.
- `color`: a `#RRGGBB` hex value (written to `res/values/colors.xml` as `cio_notification_color`), or a reference to an existing color resource (`@color/...`). Written to the manifest as `com.google.firebase.messaging.default_notification_color`.

Both meta-data entries are written as `android:resource` references, which is what the Firebase SDK reads (the existing channel metadata helper writes `android:value`, which Firebase ignores for these keys, hence the new `addResourceMetadataIfNotExists` helper).

## Behavior details

- Neither option set: no-op, current behavior unchanged.
- A meta-data entry the app already declares is never overwritten, so this composes with `expo-notifications` or manual manifest config.
- A malformed `color` fails prebuild with a clear message (matching how Live Notifications branding colors are validated).
- An `icon` path that does not resolve to a file logs a warning and skips both the copy and the manifest entry, so the resource build cannot break on a dangling `@drawable` reference.

## Testing

- New scenario tests (`__tests__/scenarios/android/notificationIconColor.test.ts`) for the pure transformers and resolvers.
- New unit tests (`__tests__/android/withNotificationIconAndColor.test.ts`) for the plugin wiring, icon copy, colors.xml write, and the skip/no-op/throw paths.
- Test app now configures both options, and `__tests__/android/app-manifest.test.js` verifies the generated manifest entries, the copied drawable, and the colors.xml entry after a real `expo prebuild`.
- Full suite verified locally per CI cell: `npm run setup-test-app && npm test -- __tests__/android` and both iOS provider cells pass; lint, `tsc --noEmit`, and the API Extractor report are updated.

Closes #115

---

> [!NOTE]
> **Low Risk**
> Build-time Expo config plugin changes only; additive with no-op when unset and safe skips for missing icons, with limited blast radius beyond Android manifest/resources.
> 
> **Overview**
> Adds **`android.pushNotification.icon`** and **`android.pushNotification.color`** so managed Expo apps can set Firebase’s default notification icon and accent color at prebuild instead of hand-editing `AndroidManifest.xml`.
> 
> A new **`withNotificationIconAndColor`** config plugin writes `com.google.firebase.messaging.default_notification_icon` and `default_notification_color` as **`android:resource`** meta-data (not `android:value`). Local icon paths are copied to `cio_notification_icon` in drawables; `#RRGGBB` colors are written to `cio_notification_color` in `colors.xml`. Existing `@drawable/` and `@color/` references pass through unchanged. Missing icon files warn and skip manifest/copy; bad colors fail prebuild; duplicate meta-data names are not overwritten.
> 
> Public types and API extractor output are updated; the test app and compatibility script configure icon/color; integration and unit/scenario tests cover manifest, resources, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ead1a80abe673aa5fc640ba6bdaa6e45d922ac8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,200 @@
...
+  'com.google.firebase.messaging.default_notification_icon';
+export const FIREBASE_NOTIFICATION_COLOR_METADATA =
+  'com.google.firebase.messaging.default_notification_color';
+
+/** Fixed name of the drawable the plugin copies a local icon image to. */
+export const NOTIFICATION_ICON_ASSET = 'cio_notification_icon';
+/** Fixed name of the color resource the plugin writes a hex color to. */
+export const NOTIFICATION_COLOR_RESOURCE = 'cio_notification_color';
+
+const HEX_COLOR = /^#[0-9a-fA-F]{6}$/;
```
mahmoud-elmorabea: Is this really always gonna be 6? Could be 8 with transparency mask?

plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,200 @@
...
+  // Android drawable names must be lowercase alphanumeric/underscore, so the asset name is fixed
+  // and only the extension follows the source file.
+  const extension = path.extname(sourcePath).toLowerCase() || '.png';
+  const drawableDir = `${androidPath}/app/src/main/res/drawable`;
+  const destination = `${drawableDir}/${NOTIFICATION_ICON_ASSET}${extension}`;
+
+  try {
+    FileManagement.mkdir(drawableDir, { recursive: true });
+    FileManagement.copyFile(sourcePath, destination);
+  } catch (error: unknown) {
```
mahmoud-elmorabea: If the copy throws we log and move on, but the caller still writes the meta-data (L182), so the manifest ends up referencing a drawable that never landed and AAPT fails later with "resource not found". The doc comment on `resolveNotificationIcon` already says the entry must be skipped when the drawable was never created — the missing-file path honors that, this path does not. Repro'd with a test where `copyFile` throws. Maybe return success from here and skip the meta-data on failure, same as the missing-file case?

plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,200 @@
...
+ */
+export const withNotificationIconAndColor: ConfigPlugin<
+  CustomerIOPluginOptionsAndroid
+> = (config, props) => {
+  const icon = props.pushNotification?.icon;
+  const color = props.pushNotification?.color
+    ? resolveNotificationColor(props.pushNotification.color)
+    : null;
+
+  if (color?.value) {
```
mahmoud-elmorabea: This runs regardless of whether the manifest mod below ends up owning the entry. If the manifest already has `default_notification_color` (bare workflow, `prebuild --no-clean`, another plugin), the old reference is kept but we still write `cio_notification_color` into colors.xml — an orphaned entry nothing references. Repro'd in a test. Could the existence check gate this too, so the value is only written when we actually own the manifest entry?

plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,200 @@
...
+): void => {
+  if (!application['meta-data']) {
+    application['meta-data'] = [];
+  }
+
+  const hasMetadata = application['meta-data'].some(
+    (metadata) => metadata.$['android:name'] === name
+  );
+
+  if (!hasMetadata) {
```
mahmoud-elmorabea: When the existing entry wins, nothing is logged — so a changed icon/color in app config is silently ignored on bare / `--no-clean` projects (the scenario tests pin this as "preserves the first resource"). Keeping the existing entry seems right, but a warn here ("already set in the manifest, leaving it — run a clean prebuild to apply") would save someone a confusing afternoon.

plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,200 @@
...
+ * Copy the icon image into `android/app/src/main/res/drawable`, where the manifest's
+ * `android:resource` reference resolves it by name.
+ */
+function installNotificationIcon(
+  sourcePath: string,
+  androidPath: string
+): void {
+  // Android drawable names must be lowercase alphanumeric/underscore, so the asset name is fixed
+  // and only the extension follows the source file.
+  const extension = path.extname(sourcePath).toLowerCase() || '.png';
```
mahmoud-elmorabea, line 158: nit: the extension is unvalidated — an `.svg` copies fine here and only fails later in AAPT. Also switching e.g. `.png` to `.webp` across non-clean prebuilds leaves both files behind, which is a duplicate-resource error. Maybe whitelist the AAPT-supported formats and remove stale siblings before copying? or we just document what formats we support

mahmoud-elmorabea (changes requested):

plugin/src/android/withNotificationIconAndColor.ts
```diff
@@ -0,0 +1,255 @@
...
+        : true;
+      if (installed) {
+        addResourceMetadataIfNotExists(
+          application[0],
+          FIREBASE_NOTIFICATION_ICON_METADATA,
+          resolvedIcon.resource
+        );
+      } else {
+        logger.warn(
+          `[customerio-expo-plugin] Skipping the ${FIREBASE_NOTIFICATION_ICON_METADATA} manifest entry because the icon could not be copied; referencing a drawable that never landed fails the Android resource build. Push notifications will use the default (grey) icon.`
```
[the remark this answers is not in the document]
mahmoud-elmorabea: This is intentional. When the config names an icon, the previous icon should not silently survive — a stale drawable masking the configured one is exactly what this section is meant to prevent. The window here needs a non-clean prebuild, an extension change, and a copy failure all at once, and the result is a loud AAPT error rather than shipping the wrong icon, which is the trade we want.

mahmoud-elmorabea (approved):

PR #398, opened 2026-08-26
feat: support Expo scene lifecycle routing (MBL-2304)

Shahroz16:
## Summary

- preserve Expo SDK 58 scene ownership while installing Customer.io push and SDK-originated deep-link routing before native auto-initialization
- move URL ownership only when the scene manifest and executable `SceneDelegate: ExpoAppSceneDelegate` implementation confirm Expo owns the lifecycle
- use Expo Router's released `+native-intent` contract for Live Activity attribution and destination replacement
- keep generated host, notification-service, Live Activity, and Podfile reconciliation across APN/FCM and linkage changes
- keep the scene-routing checks deterministic and document the one customer-owned JavaScript integration point

## Dependency

React Native #646 and #647 are merged and released in `customerio-reactnative` 6.9.0. This PR and its test app now use that published version. Expo #49391 is not a release gate, and no branch or commit SHA ships as a package dependency.

## Validation

- focused scene and warning tests, TypeScript, lint, and package build pass locally
- React Native's merged Xcode 27 scene E2E passes with the real Live Activity bridge, including invalid-URL passthrough and JavaScript `null` mapping
- an exact packed stack generated, built, installed, and launched a real Expo 58 scene host; Maestro verified warm and cold Live Activity destination routing through `+native-intent`
- native iOS tests separately verify opened attribution and destination extraction
- Expo CI is validating the final published-version stack

Part of MBL-2304.


plugin/src/ios/utils.ts
```diff
@@ -26,6 +221,16 @@ export const isExpoVersion53OrHigher = (config: ExpoConfig): boolean => {
...
 };
 
+/** Returns true if Expo uses the generated UIScene lifecycle (SDK >= 58.0.0). */
+export const isExpoVersion58OrHigher = (config: ExpoConfig): boolean => {
+  // Expo prereleases already use the next SDK's native template. Compare the coerced major version
+  // here without changing prerelease behavior for the older version gates above.
+  const coercedVersion = semver.coerce(config.sdkVersion || '');
+  return coercedVersion
+    ? semver.gte(coercedVersion, '58.0.0')
+    : false;
```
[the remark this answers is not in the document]
Shahroz16: This is not how semver.coerce behaves. It normalizes 58.0.0-beta, 58.0.0-rc, and canary strings to 58.0.0 before the gte comparison. Focused prerelease cases pass and the generated Expo 58 canary entered the scene path during validation, so no code change is needed.

plugin/src/ios/withCIOIos.ts
```diff
@@ -141,6 +201,10 @@ export function withCIOIos(
     config = withGeofenceAppDelegate(config);
   }
 
+  if (usesSceneLifecycle && liveNotificationsEnabled) {
+    config = withLiveActivityUrlRoutingWarning(config);
```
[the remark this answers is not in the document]
Shahroz16: This hook was never released in any Expo plugin version; it existed only on the earlier draft of this same PR. Published customers therefore cannot have a Customer.io-generated transformURL block to migrate. Shipping permanent source-rewrite cleanup for unreleased draft code would add maintenance risk and could touch customer-owned SceneDelegate code, so no cleanup is added.

plugin/src/ios/utils.ts
```diff
@@ -1,7 +1,199 @@
...
+  const infoPlist = fs.readFileSync(infoPlistPath, 'utf8');
+  const sceneDelegateClassName = infoPlist
+    .match(
+      /<key>UISceneDelegateClassName<\/key>\s*<string>([^<]+)<\/string>/
+    )?.[1]
+    ?.trim();
+  return (
+    infoPlist.includes('<key>UIApplicationSceneManifest</key>') &&
+    sceneDelegateClassName === '$(PRODUCT_MODULE_NAME).SceneDelegate'
+  );
```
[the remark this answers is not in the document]
Shahroz16: You are right that the manifest and filename alone do not prove Expo owns the scene lifecycle. Commit 557d6c1 now inspects executable SceneDelegate source and requires SceneDelegate to inherit ExpoAppSceneDelegate before moving URL ownership away from AppDelegate. Otherwise the plugin keeps the AppDelegate path. The regression test covers a custom non-Expo delegate and a commented fake declaration.

plugin/src/ios/withCIOIosSwift.ts
```diff
@@ -365,18 +484,63 @@ export function modifyAppDelegateForLiveActivityUrl(contents: string): string {
...
+  if (hasExecutableSceneRouting(contents)) {
+    return contents;
+  }
+
+  const executableContents = maskSwiftNonCode(contents);
+  const initializationMatch =
+    executableContents.match(PUSH_INITIALIZATION_LINE_REGEX) ??
+    executableContents.match(NATIVE_INITIALIZATION_LINE_REGEX);
+  if (!initializationMatch || initializationMatch.index === undefined) {
+    throw new Error(
```
mrehan27, line 521: **P2** — this throws, aborting the prebuild, and the precondition is an exact line match.

`PUSH_INITIALIZATION_LINE_REGEX` (:44) requires exactly `cioSdkHandler.application(application, didFinishLaunchingWithOptions: launchOptions)` with only surrounding whitespace. `_ = cioSdkHandler.application(...)` or a trailing comment won't match.

That bites hardest on the upgrade path: `modifyAppDelegateForPushHandler` reaches here from the "changes already exist" branch, whose own comment notes an older plugin version's AppDelegate may differ. Every other unmet scene precondition in this file warns and falls back to AppDelegate routing — worth matching that here.

plugin/src/ios/withLiveActivityUrlRoutingWarning.ts
```diff
@@ -0,0 +1,40 @@
+import { withDangerousMod } from '@expo/config-plugins';
+import type { ExpoConfig } from '@expo/config-types';
+import fs from 'fs';
+import path from 'path';
+import { logger } from '../utils/logger';
+
+const NATIVE_INTENT_EXTENSIONS = ['tsx', 'ts', 'jsx', 'js'];
+
+export function hasExpoRouterNativeIntent(projectRoot: string): boolean {
```
mrehan27, line 9: **P2** — this checks only that a `+native-intent` file exists, not that it calls `handleWidgetUrl`. An app whose native-intent file was written for something else gets no warning, which is the case most likely to ship with broken Live Activity attribution.

Grepping that file for `handleWidgetUrl` would make the check mean what the warning says. The extension list is correct, incidentally — it matches expo-router's own `/^\.\/\+native-intent\.[tj]sx?$/`.

plugin/src/ios/withCIOIosSwift.ts
```diff
@@ -277,36 +354,66 @@ export const withCIOIosSwift = (
   }
 };
 
+function warnIfNativeAutoInitializationNeedsSceneReadiness(
```
mrehan27: **P2** — this fires on every prebuild of an auto-init scene app, and a correctly configured app can't silence it, since the plugin can't see that the JS already calls `setDeepLinkRoutingReady()`.

With the native-intent warning that's two permanent prebuild warnings surviving correct setup, which is how warnings stop being read. Detecting the call in JS source, or emitting only on configuration change, would keep the signal meaningful.

mrehan27 (approved):
Approving. Dropping the unreleased `transformURL` integration in favour of Expo Router's released `+native-intent` contract is a real de-risking — no upstream dependency left to gate the design.

Two things improved beyond the refactor: `maskSwiftNonCode` is now applied consistently to presence checks, so a class name in a comment no longer reads as "already integrated"; and the Podfile helpers moved from skip-if-present to replace-managed-block, which fixes provider and linkage changes being ignored on incremental prebuilds.

All three Bugbot findings are resolved — I verified the semver one is a false positive as you said, and that `withCIOSceneDelegate.ts` never existed outside this PR.

Still blocked on a published React Native release containing customerio/customerio-reactnative#646 and customerio/customerio-reactnative#647.

plugin/src/ios/withCIOIosSwift.ts
Shahroz16: Partly agreed. Trailing comments were already safe because matching uses masked Swift source, but the discard-assignment form was too strict. Commit 5c23cd4 accepts optional _ =, host variable names, whitespace, and semicolons, with an upgrade regression test. I kept the hard failure when no safe executable anchor exists because falling back to AppDelegate routing in a confirmed scene-owned host would silently miss scene URL delivery.

plugin/src/ios/withLiveActivityUrlRoutingWarning.ts
Shahroz16: Agreed. Commit 5c23cd4 now suppresses the warning only when +native-intent contains the CustomerIO.liveActivities.handleWidgetUrl call, and covers an unrelated native-intent file. Commit 7928b30 further limits detection to executable call forms so comments and example strings do not produce a false positive.

plugin/src/ios/withCIOIosSwift.ts
Shahroz16: Agreed on the warning fatigue. Commit 5c23cd4 demotes this to informational output. I did not add recursive JavaScript scanning because source presence cannot prove that the Linking listener was registered before the readiness call and could falsely silence the diagnostic. The required ordering remains explicit in the README.

plugin/src/ios/withLiveActivityUrlRoutingWarning.ts
```diff
@@ -0,0 +1,47 @@
...
+      const nativeIntentPath = path.join(
+        projectRoot,
+        appRoot,
+        `+native-intent.${extension}`
+      );
+      if (!fs.existsSync(nativeIntentPath)) return false;
+
+      return /\bCustomerIO\s*\.\s*liveActivities\s*\.\s*handleWidgetUrl\s*\(/.test(
+        fs.readFileSync(nativeIntentPath, 'utf8')
+      );
```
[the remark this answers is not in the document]
Shahroz16: Valid. Commit 7928b30 anchors detection to recognizable executable call forms and adds a regression case where the helper appears only in a comment and string. That keeps the check narrow without introducing a JavaScript parser into the config plugin.

plugin/src/ios/withLiveActivityUrlRoutingWarning.ts
Shahroz16: Follow-up after implementing this: the next review correctly showed that a regex capable of distinguishing executable JavaScript from comments also rejects valid call shapes such as return await, typed assignments, aliases, and composed expressions. Proving semantic routing here would require a real JS/TS parser and still would not prove runtime ordering. I rolled the content grep back in e052553. The plugin now keeps the deterministic check for the documented +native-intent integration point, while the README and scene E2E own the helper contract. This avoids turning a prebuild diagnostic into fragile source analysis.

plugin/src/ios/withLiveActivityUrlRoutingWarning.ts
```diff
@@ -0,0 +1,47 @@
...
+  return ['app', path.join('src', 'app')].some((appRoot) =>
+    NATIVE_INTENT_EXTENSIONS.some((extension) => {
+      const nativeIntentPath = path.join(
+        projectRoot,
+        appRoot,
+        `+native-intent.${extension}`
+      );
+      if (!fs.existsSync(nativeIntentPath)) return false;
+
+      return /^[ \t]*(?:(?:return|await)[ \t]+|(?:const|let|var)[ \t]+[A-Za-z_$][A-Za-z0-9_$]*[ \t]*=[ \t]*(?:await[ \t]+)?)?CustomerIO\s*\.\s*liveActivities\s*\.\s*handleWidgetUrl\s*\(/m.test(
```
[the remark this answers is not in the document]
Shahroz16: Valid, and it confirms the broader issue with parsing customer JavaScript using a regex. Commit e052553 removes the content inspection instead of expanding it into a growing list of valid syntax forms. The check now only detects the documented +native-intent integration point; the required helper remains explicit in the README and is exercised by the scene E2E.

PR #399, opened 2026-08-29
ci: add fast and full PR validation lanes

Shahroz16:
## Summary
- keep representative Android and iOS/FCM real-build compatibility coverage on routine PRs
- add the redundant iOS/APN cell and sample distribution only with `ci:full`
- scope the expensive pnpm native compatibility workflow to relevant files
- remove duplicate feature-branch push runs while preserving main and beta coverage

## Validation
- workflow tests passed (2/2)
- scenario tests passed (23 suites, 151 tests, 63 snapshots)
- API Extractor and package build passed
- changed workflow YAML passes differential Actionlint validation
- no release workflow changed

### DIFF
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -55,9 +55,15 @@ while IFS= read -r -d '' commit_message; do
   fi
 done < "$commit_messages_file"
 
+full_validation=false
+requires_full_label=false
+
+if [[ "$pr_release" == 'true' || "$commit_release" == 'true' || "${CI_FULL:-false}" == 'true' ]]; then
+  full_validation=true
+fi
+
 if [[ "$commit_release" == 'true' && "$pr_release" != 'true' && "${CI_FULL:-false}" != 'true' ]]; then
-  echo "::error::This PR contains a release-bearing commit, but its PR title/body is non-release. Retitle it with fix:, feat:, perf:, revert:, or a breaking-change marker, or apply ci:full, so release validation runs before merge."
-  exit 1
+  requires_full_label=true
 fi
 
 if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
@@ -65,4 +71,17 @@ if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSI
   exit 1
 fi
 
-echo "Release validation policy satisfied (pr_release=$pr_release, commit_release=$commit_release)."
+if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
+  {
+    echo "pr_release=$pr_release"
+    echo "commit_release=$commit_release"
+    echo "full_validation=$full_validation"
+    echo "requires_full_label=$requires_full_label"
+  } >> "$GITHUB_OUTPUT"
+fi
+
+if [[ "$requires_full_label" == 'true' ]]; then
+  echo "::notice::A release-bearing commit was detected behind a non-release PR title. The semantic PR helper will apply ci:full automatically."
+fi
+
+echo "Release validation policy satisfied (pr_release=$pr_release, commit_release=$commit_release, full_validation=$full_validation)."
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -66,7 +66,9 @@ if [[ "$commit_release" == 'true' && "$pr_release" != 'true' && "${CI_FULL:-fals
   requires_full_label=true
 fi
 
-if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
+if [[ "${ENFORCE_TRUSTED_RELEASE:-true}" == 'true' &&
+  ("$pr_release" == 'true' || "$commit_release" == 'true') &&
+  ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
   echo "::error::Release-bearing PRs require secret-dependent validation from a trusted repository branch. Recreate or update this change on an internal branch before merge."
   exit 1
 fi
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${PR_TITLE:?PR_TITLE is required}"
+: "${BASE_SHA:?BASE_SHA is required}"
+: "${HEAD_SHA:?HEAD_SHA is required}"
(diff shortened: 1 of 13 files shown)


Shahroz16:
## Revised CI decision after correctness review

Exact head: `4d78425`

The review feedback was correct: semantic-release intent is not a correctness boundary. A `chore:`, `ci:`, or native dependency PR can break Expo prebuild, Android, APN, FCM, or a supported Expo version. This head keeps those checks automatic.

- Every code PR runs the latest Expo Android and iOS APN native smoke pair.
- Compatibility-sensitive native, dependency, script, app, and workflow paths automatically add four non-overlapping cells: Expo 54 Android, APN, and FCM, plus latest Expo FCM.
- Main, release, and manual validation retain the full supported six-cell matrix.
- Java setup runs only on Android jobs.
- PR title and commit semantics control only signed sample distribution side effects.
- Stable SDK release and deploy workflows are unchanged.

This CI-changing PR intentionally exercised all six compatibility cells with no `ci:full` label. Final hosted result: **20 successful, 1 intentional skip, 0 failures**. Signed sample distribution skipped as designed.

The four path-gated delta cells consumed **66m22s of Mac runner time plus 21m30s on Linux** on this final run. Routine changes retain the latest Android and APN smoke pair while avoiding those four cells; compatibility-sensitive changes still run all six automatically.

Claude Code Opus rejected the earlier semantic lane split, then returned **APPROVE** after the path-based redesign. All changed workflows pass actionlint and the final hosted head is green.

mrehan27 (commented):
Cross-repo review — the same change across five repos.

The lane gating looks right: omitting `always()` so a skipped `prepare` transitively blocks E2E, and moving the dependabot guard to `github.event.pull_request.user.login` (on a `labeled` event `github.actor` is the labeler, not the author).

Main ask: the release-policy script and reusable workflow are byte-identical in all five repos today (sha256 match), so extracting them to `mobile-ci-tools` is free now and a five-repo change after merge.

.github/scripts/validate-release-policy.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
mrehan27, line 1: **P2** — this script and `reusable-release-policy.yml` are byte-identical across all five SDK repos (sha256 match), with nothing keeping them in sync. `customerio/mobile-ci-tools` already hosts shared reusable workflows — this change calls one. Moving it there is mechanical with zero drift today; after merge, every policy fix becomes a five-repo change.

.github/workflows/validate-plugin-compatibility-full.yml
```diff
@@ -0,0 +1,35 @@
...
+      target_repository: ${{ github.repository }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      ci_full: ${{ contains(github.event.pull_request.labels.*.name, 'ci:full') }}
+
+  # Metadata edits rerun full so a newer skipped check cannot replace a failure.
+  full-compatibility:
+    needs: release_policy
+    if: >-
+      always() &&
+      (github.event.action == 'edited' ||
```
mrehan27: **P2** — `github.event.action == 'edited' || github.event.action == 'labeled'` forces the full lane regardless of classification, so a title typo fix or any unrelated label runs the expensive lane on a non-release PR. Gating `labeled` on `github.event.label.name == 'ci:full'` keeps the intent without that case.

.github/workflows/build-sample-apps-full.yml
```diff
@@ -0,0 +1,39 @@
+name: Full sample app builds
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited, labeled]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
mrehan27, line 19: **P2** — concurrency policy diverges across the five repos:

- android: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`
- ios, flutter: same, plus `|| startsWith(github.ref, 'refs/heads/feature/')`
- reactnative: ungated `cancel-in-progress: true`
- expo-plugin: both forms, in the same PR

Worth converging, or a comment if the differences are deliberate.

.github/workflows/validate-plugin-compatibility-full.yml
Shahroz16: Addressed on current head `b6fb5dd6`. A bare metadata event no longer requests the full lane: release classification still controls eligibility, `ci:full` must be the label that fired the labeled event, and unrelated labels/body edits are isolated in the `metadata` concurrency group.

.github/scripts/validate-release-policy.sh
Shahroz16: Agreed that five copies create drift risk. I am deliberately not extracting this classifier in these PRs because it gates signed/published side effects, and moving it immediately to the existing mutable shared-workflow dependency would enlarge the blast radius of a policy change. The safe follow-up is to version and pin the shared policy, canary one SDK, then migrate the other callers. For this rollout the local copies are byte-identical and fixture-tested.

.github/workflows/build-sample-apps-full.yml
Shahroz16: Reviewed across all five SDKs. The PR cancellation contract is consistent, including the new `metadata` versus `validation` isolation. The remaining `cancel-in-progress` syntax differs because non-PR trigger scope differs: iOS/Flutter retain cancellation for feature-branch pushes, Android preserves non-PR sample/release runs, Expo's full wrapper is PR-only, and React Native preserves scheduled/manual/reusable E2E. Converging the surface syntax here would change existing non-PR behavior.

.github/workflows/build-sample-apps-full.yml
```diff
@@ -0,0 +1,46 @@
...
+  # Unrelated labels and body-only edits must not cancel an in-flight release or
+  # ci:full run. They share a separate group from meaningful validation events.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      github.event.changes.title.from == '' &&
+      github.event.changes.base.ref.from == '')) &&
+      'metadata' || 'validation'
```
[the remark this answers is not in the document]
Shahroz16: Addressed in `e763d6f`. The concurrency predicate now checks for the presence of `github.event.changes.title` / `.base` objects directly. Any title-change payload is therefore `validation` regardless of its prior string value; body-only edits remain `metadata`. `actionlint` passes and Opus approved the explicit truth table.

mahmoud-elmorabea (approved):
- **`build-sample-apps-full.yml` has no `paths` / `paths-ignore` filter**, so a markdown-only release-bearing PR triggers full signed sample distribution across Android + iOS APN + iOS FCM. Every other workflow in this PR gained a path filter; this one should too.

Shahroz16:
I checked the markdown-only suggestion against the release contract and intentionally did not add a trigger-level paths-ignore filter.\n\nA markdown-only fix or revert can still be release-bearing under semantic-release. Filtering the workflow before classification would let that release path bypass signed sample validation and would also make the ci:full escape hatch unreachable. The current job gate already skips signed builds for non-release documentation PRs; only the lightweight Linux classifier runs.\n\nThat preserves the optimization for ordinary documentation changes without creating a publish-without-validation path.

PR #400, opened 2026-08-31
chore: resolve the published SDK version from the release job

mahmoud-elmorabea:
Fixes MBL-2324

## Problem

`reusable_build_sample_apps.yml` resolved the SDK version to build the test apps against with `git describe --tags --abbrev=0`, evaluated on the **pre-release** commit.

semantic-release tags a **new child commit** of that commit. `git describe` only walks ancestors, so on the release path it returns the *previous* release. That value became `sdkVersion=` in `test-app/local.env`, and `scripts/applyLocalEnvValues.js` → `updateSdkVersion()` installed that stale plugin version.

Confirmed in [run 33082901972](https://github.com/customerio/customerio-expo-plugin/actions/runs/33082901972): the release commit raised the test app's `customerio-reactnative` constraint to `^6.9.0`, but CI installed plugin **3.8.0**, whose exact peer is `6.6.3` → `npm ERESOLVE`. Not a publish race — 3.9.0 was on npm at 14:36:10Z and the failing install ran at 14:37:34Z. The job simply asked for the wrong version. Same fingerprint in four earlier runs.

## Fix

`LATEST_TAG` was answering two different questions with one value. They are now separated:

| value | question | source |
| --- | --- | --- |
| `LATEST_TAG` | "newest tag reachable from this commit" — an **ancestry** question | `git describe` (unchanged) |
| `SDK_BUILD_VERSION` | "which published SDK do these apps install" | new `sdk_version` input, falling back to `LATEST_TAG` |

- `reusable_build_sample_apps.yml` gains an optional `sdk_version` input. When it is non-empty it wins; otherwise the job falls back to `git describe`.
- On a **release** build (`use_latest_sdk_version: true`) that is *not* a `workflow_dispatch`, an empty input is now a hard error rather than a silent fallback — that silent fallback is exactly what shipped the wrong version.
- `commitsAheadCount` still reads `LATEST_TAG`. `git describe` is the correct answer there and restricting it would have broken every PR and push build.
- `SDK_BUILD_VERSION` also drives the two other "which version is this app" consumers — the `update_expo_test_app_version` fastlane stamp and the `sdk_version` field in the sample-app Slack notification — which were reporting the stale tag for the same reason. On non-release builds the input is empty, so both resolve to `LATEST_TAG` exactly as before; nothing changes for PR/push runs.
- `build-sample-app-for-sdk-release.yml` accepts and forwards `sdk_version` on both `workflow_dispatch` and `workflow_call`.
- `deploy-sdk.yml`: `needs: deploy-npm` → `needs: [deploy-git-tag, deploy-npm]`, so `needs.deploy-git-tag.outputs.new_release_version` is reachable. `deploy-git-tag` already declared that output; `needs` exposes outputs of directly-listed jobs only, so this is purely about reach — the ordering was already guaranteed transitively through `deploy-npm`, and no job gating changes.

The deploy path is a nested reusable-workflow `uses:` chain (`deploy-sdk` → `build-sample-app-for-sdk-release` → `reusable_build_sample_apps`), so `github.event_name` inside the reusable workflow is the originating event. A manual dispatch of `build-sample-app-for-sdk-release.yml` therefore still reports `workflow_dispatch` and keeps its `git describe` fallback, which is correct there — no release is being cut concurrently.

Reference implementation: `customerio-reactnative`'s `build-release-sample-apps.yml` (read only; unchanged).

## Notes / risk

- Behaviour on PR and push builds is byte-identical: the input is empty, so `SDK_BUILD_VERSION == LATEST_TAG`.
- The exact `customerio-reactnative` peer pin in `package.json` is deliberately **untouched**. `plugin/src/utils/writeExpoVersion.ts` writes into `node_modules/customerio-reactnative/package.json` and `plugin/src/helpers/utils/injectCIOPodfileCode.ts` resolves that SDK's pods by `:path`; 25+ commits move it in lockstep. The bug was CI asking for the wrong version, not the pin.
- Verified with `actionlint` (clean apart from pre-existing `SC2086:info` on unquoted `$GITHUB_ENV`, which matches the file's existing style).

### DIFF
--- a/.github/workflows/build-sample-app-for-sdk-release.yml
+++ b/.github/workflows/build-sample-app-for-sdk-release.yml
@@ -2,11 +2,23 @@ name: Publish test apps for SDK release
 
 on:
   workflow_dispatch:
+    inputs:
+      sdk_version:
+        description: "Published plugin version to build the test apps against (e.g. 3.9.0). Leave empty to use the newest tag reachable from this commit."
+        required: false
+        type: string
   workflow_call:
+    inputs:
+      sdk_version:
+        description: "Published plugin version to build the test apps against. The deploy workflow passes the version semantic-release just tagged."
+        required: false
+        type: string
+        default: ""
 
 jobs:
   build-sample-apps:
     uses: ./.github/workflows/reusable_build_sample_apps.yml
     with:
       use_latest_sdk_version: true
+      sdk_version: ${{ inputs.sdk_version }}
     secrets: inherit
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -294,6 +294,12 @@ jobs:
         run: npm publish
 
   publish-sample-apps-public-builds:
-    needs: deploy-npm
+    # `deploy-git-tag` is needed for its `new_release_version` output, not for
+    # ordering — `needs` only exposes the outputs of directly-listed jobs.
+    # Without it the test apps resolve their version with `git describe`, which
+    # walks ancestors only and so names the PREVIOUS release on a release commit.
+    needs: [deploy-git-tag, deploy-npm]
     uses: ./.github/workflows/build-sample-app-for-sdk-release.yml
+    with:
+      sdk_version: ${{ needs.deploy-git-tag.outputs.new_release_version }}
     secrets: inherit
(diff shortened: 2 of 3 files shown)


Shahroz16 (approved):
Core version-resolution change looks good and fixes the release mismatch. One small hardening follow-up: the manual `sdk_version` value is interpolated directly into shell commands in the reusable workflow. I would validate it as a release version and use `$SDK_BUILD_VERSION` in the command instead. Since only trusted repo users can dispatch this workflow, I do not think it needs to block this PR.

PR #401, opened 2026-08-31
chore: raise the Android testbed Gradle heap and fix its cache key

mahmoud-elmorabea:
Fixes MBL-2324

> Stacked on #400. Targets `mbl-2324-expo-release-version-resolution`; both changes are in `reusable_build_sample_apps.yml`. Review only [the second commit](https://github.com/customerio/customerio-expo-plugin/pull/401/commits).

Two independent defects in the Android testbed job.

## 1. dex-merge OOM

[Run 33078507086](https://github.com/customerio/customerio-expo-plugin/actions/runs/33078507086) died with `D8: java.lang.OutOfMemoryError: Java heap space` during `:app:mergeDexRelease`.

Verified against the generated project:

- `test-app/android/gradle.properties` sets `org.gradle.jvmargs=-Xmx2048m -XX:MaxMetaspaceSize=512m`.
- `test-app/android/app/build.gradle:70` defaults `enableMinifyInReleaseBuilds` to `false`, and nothing sets `android.enableMinifyInReleaseBuilds` — so the merge is over an unshrunk graph.
- D8 runs **in-process** in the daemon, so that 2048m cap is the ceiling for the merge.

That `gradle.properties` is generated by `expo prebuild` and gitignored, so it cannot be edited. The plugin's `withGradleProperties` mods are also not an option — those ship to customers, and this is a CI-only constraint.

Gradle ranks `GRADLE_USER_HOME/gradle.properties` above the project's, so the fix is an append there — the same pattern `customerio-flutter` already uses for the identical constraint in `reusable-build-sample-apps.yml`, and the same `4g`:

```yaml
- name: Increase Gradle memory limit
  run: |
    mkdir -p ~/.gradle
    echo "org.gradle.jvmargs=-Xmx4g" >> ~/.gradle/gradle.properties
```

`_JAVA_OPTIONS: "-Xmx4g"` is mirrored on the fastlane step's `env`, as Flutter does.

Two things checked deliberately:

- **The append stays clean per run.** The Gradle cache action restores only `~/.gradle/caches` and `~/.gradle/wrapper`; `~/.gradle/gradle.properties` is under neither, so it is never restored and the `>>` cannot accumulate across runs.
- `mkdir -p ~/.gradle` is the one deviation from the Flutter step. Flutter's runs after `setup-android`, which has already created the directory; here the memory step can run on a cache miss with no `~/.gradle` present, and a bare `>>` would fail.

Note that Gradle resolves `org.gradle.jvmargs` per-property, not by merging, so the override also drops `-XX:MaxMetaspaceSize=512m`. That is the same trade Flutter makes; metaspace was not the failure mode and an unbounded default is the safer side.

## 2. Wrong Gradle cache key

The key was `hashFiles('package-lock.json')` — the **root** lockfile, which pins the *plugin's* devDependencies. `test-app`'s own lockfile is not committed, and the root one does not move when the test app's dependency graph moves. Four runs shared a single cache entry while building three different `customerio-reactnative` versions. The in-code comment claimed the key invalidated "when plugin/test-app deps change", which is the opposite of what it did.

Re-keyed on the two inputs that actually vary — `test-app/package.json` and the resolved SDK version (`SDK_BUILD_VERSION`, from #400) — with a two-level `restore-keys` ladder so a version bump still gets a warm start. The step moved below `Capture Git Context` because it now consumes that value. Comment rewritten to describe what the key does.

## Out of scope, flagged

The **CocoaPods** cache in the iOS job (`reusable_build_sample_apps.yml`) has the identical defect and the identical misleading comment — it keys on the same root `package-lock.json` while `test-app/Podfile.lock` is gitignored. Left alone here to keep this PR matching its title and to keep the stack reviewable; worth a follow-up.


Shahroz16 (approved):
Looks good to me. The cache key now varies with the dependency inputs, and the 4 GB Gradle heap is applied effectively for the testbed build.

PR #402, opened 2026-08-31
chore: make the Expo test app dependency graph reproducible

mahmoud-elmorabea:
Fixes MBL-2324

> **Highest-risk change in the MBL-2324 batch — it alters how every Expo CI build resolves dependencies.** Opened as a draft for review; not expected to merge soon. Stacked on #401 → #400. Review only [the third commit](https://github.com/customerio/customerio-expo-plugin/pull/402/commits).

## Problem

`test-app/.gitignore` ignored `package-lock.json`, and `scripts/setup-test-app.sh` ran a plain `npm install` against floating ranges (`~55.0.x`, `^6.9.0`). Three CI runs resolved three different `customerio-reactnative` versions (6.7.0 / 6.9.0 / 6.10.0) in two days. Two runs of the same commit could build different apps, which makes every red build ambiguous.

`npm ci` was impossible because `scripts/applyLocalEnvValues.js` ran as the test app's `preinstall` hook and rewrote `dependencies` in `test-app/package.json` at install time — `npm ci` rejects a manifest that moves under it.

## Approach

**The lockfile owns the static graph; exactly one edge is deliberately floating.**

1. `test-app/package-lock.json` is un-ignored and committed (709 packages).
2. `test-app/package.json` drops the `preinstall` hook and the `customerio-expo-plugin` dependency. That dependency was `file:../customerio-expo-plugin-latest.tgz`, whose integrity hash changes every time the plugin is rebuilt — leaving it in the lockfile would break `npm ci` on the next commit.
3. `applyLocalEnvValues.js` no longer touches `package.json`. It keeps the `app.json` push-provider update and gains a `--print-sdk-version` mode so it stays the single parser of `local.env`, for the shell as well.
4. `setup-test-app.sh` becomes: apply app config → `npm ci` → install the plugin explicitly → `expo prebuild`.
5. `install-plugin-tarball.sh` installs with `--no-save`.
6. The Gradle cache key from #401 moves from `test-app/package.json` to the now-committed `test-app/package-lock.json`.

## The one edge that still moves, on purpose

`npm install customerio-expo-plugin@$SDK_VERSION --save-exact` runs **after** `npm ci`. The plugin declares an *exact* peer on `customerio-reactnative`, so that install may also move `customerio-reactnative` within its `^6.9.0` range. That is intended and unavoidable — it is the coupling the exact peer pin exists to enforce — but it is worth stating plainly: **the lockfile pins `customerio-reactnative`, and the plugin's peer can override that pin.** No other package can move after `npm ci`.

The lockfile is committed at `customerio-reactnative@6.9.0`, matching the peer the plugin declares at HEAD, so the common case is a no-op rather than a downgrade on every run. A fresh resolve of `^6.9.0` gives 6.10.0; committing that would have meant every single run correcting it.

## Verified locally, not just reasoned about

| check | result |
| --- | --- |
| `npm install` in `test-app` generates the lockfile | 709 packages, `lockfileVersion` 3, no `file:` refs, no `customerio-expo-plugin` entry |
| `npm ci` against the committed lockfile | clean, 709 packages |
| published path: `npm install customerio-expo-plugin@3.9.0 --save-exact` over a lock pinning `customerio-reactnative@6.10.0` | succeeds; npm moves RN to 6.9.0 to satisfy the exact peer — **no ERESOLVE** |
| local path: `install-plugin-tarball.sh` with `--no-save` | plugin present in `node_modules` at 3.9.0; `package.json` and `package-lock.json` byte-identical afterwards |
| full `setup-test-app.sh`, both branches (prebuild stubbed) | both complete |
| `npx jest __tests__/utils __tests__/workflows` | 184 passed |
| `actionlint` | no new findings |

CI on this PR exercises the real thing: `test.yml`'s `test-plugin-android` / `test-plugin-ios` jobs run `npm run setup-test-app` including a real `expo prebuild`, and snapshot the generated native projects.

## Collateral changes, all forced by the above

- `scripts/clean-all.sh` no longer does `rm -f test-app/package-lock.json` — that file is committed now, and deleting it would silently drop the pinned graph. (`rm -f package-lock.json` for the *root* lockfile is left alone: it is pre-existing and equally odd, but out of scope here.)
- `scripts/build-all.sh` called `npm run preinstall && npm install` and then prebuilt separately; it now delegates to `setup-test-app.sh` so the two cannot drift. This also drops a pre-existing `npx npx expo prebuild` typo.
- `local-development-readme.md`: `npm run setup-test-app` is documented as the supported entry point, and the removed `npm run preinstall` instruction is replaced.

## Honest risk read

- **Highest risk is the peer/lockfile coupling described above.** If a future plugin release raises its exact peer *outside* `^6.9.0` (a `customerio-reactnative` major), the install step fails with ERESOLVE and `test-app/package.json`'s range must be bumped in the same change. Today's `npm install` hid that by resolving whatever the range allowed; it will now be a visible, loud failure. I consider that an improvement, but it is a new failure mode.
- **A bare `npm install` in `test-app` is no longer the supported path** — it re-resolves the static graph and does not install the plugin at all. The readme is updated; anyone with muscle memory will be briefly surprised.
- **The lockfile is now a maintenance surface.** Dependabot and Expo SDK bumps have to touch it, and a stale lockfile silently means a stale test app rather than a fresh resolve.
- `setup-test-app.sh` requires the **root** dependencies to be installed first, because `applyLocalEnvValues.js` resolves `dotenv` relative to `scripts/`. This is pre-existing (the old `preinstall` hook had the identical requirement) and every CI job already runs a root `npm ci` first — flagging it because the failure message (`Cannot find module 'dotenv'`) is unhelpful, and I hit it in a fresh worktree.
- `--save-exact` for the published path but `--no-save` for the tarball path is a deliberate asymmetry, not an oversight: the published version is a real pin worth recording in the (throwaway CI) manifest, whereas a locally built tarball must never be able to reach the committed lockfile.


Shahroz16 (changes requested):
One thing I think we need to fix before merging: the committed lock graph is reproducible, but it is not Expo-coherent. From a clean install on this head, `npx expo install --check` exits 1 because React Native is 0.83.6 instead of the 0.83.10 version expected by Expo 55.0.31, and Reanimated is 4.2.3 instead of 4.2.1. That means we are freezing a graph Expo itself flags as incompatible. Can we run `expo install --fix` and regenerate the lockfile, or otherwise pin the dependency set Expo 55 expects?

Not blocking on this second point, but `BuildMetadata` should also prefer the installed `customerio-reactnative/package.json` over lockfile metadata. A future `--no-save` plugin install can change the installed RN version while the dashboard still reports the stale lockfile value.

mahmoud-elmorabea:
Both points addressed — thanks, the first one was a real hole.

**1. Expo coherence.** Reproduced exactly what you saw:

```
react-native@0.83.6           - expected version: 0.83.10
react-native-reanimated@4.2.3 - expected version: 4.2.1
```

You're right that a reproducible graph isn't worth much if Expo itself flags it as incompatible. Regenerated with `expo install --fix` and re-verified from a clean tree — `rm -rf node_modules && npm ci` then `expo install --check` now reports **Dependencies are up to date**. Lockfile churn is −188/+62.

One thing worth a look in review: `react-native-reanimated` moved from `~4.2.1` to an exact `4.2.1`, which is what `--fix` does. That caret range is precisely what let it drift to 4.2.3, so pinning fits this PR's purpose — but it is a semantics change, not just a version bump.

**2. `BuildMetadata`.** Agreed, and this was a genuine flaw in the earlier fix — I had deliberately left `customerio-reactnative` on the lockfile path. It now takes the version from the installed manifest and consults the lockfile only for the `file:` marker that drives the `-as-source` suffix, so a plugin install that moves RN within its range no longer reports a stale version.

test-app/helpers/BuildMetadata.js
```diff
@@ -57,17 +57,29 @@ function formatBuildDateWithRelativeTime(timestamp) {
...
+      );
       return undefined;
     }
 
-    const version = resolveValidOrElse(sdkPackage.version);
-    const isPathDependency =
-      sdkPackage.resolved && sdkPackage.resolved.startsWith('file:');
-    if (isPathDependency) {
+    const version = resolveValidOrElse((installed || lockEntry).version);
+    const isBuiltFromSource = Boolean(
```
mrehan27, line 79: **P2** — the `file:` marker can no longer appear: the plugin is installed `--no-save` on both paths and is gone from `package.json`, so the committed lockfile has no `customerio-expo-plugin` entry and no `file:` ref at all, leaving `isBuiltFromSource` permanently false and a local tarball build rendering identically to a published one. `extras.commitsAheadCount` is already read here, so driving the suffix off that — or off which install branch ran — would restore the distinction.

mrehan27 (approved):
The lockfile approach is right, and the `expo install --fix` pass makes the graph coherent rather than merely frozen — that was the important correction.

One item inline: the `-as-source` suffix can no longer render, so the lockfile is no longer able to supply the `file:` marker it is still being consulted for.

Shahroz16 (approved):
This addresses my earlier concerns. I reran the clean install and Expo compatibility check with both the local tarball and published plugin. Both pass without changing the committed lockfile, and the app bundles successfully.

Agree with Rehan on keeping the source-build suffix. Also worth following up on the package.json import: Metro currently resolves it through a fallback and logs a warning. Neither needs to hold up this CI fix.

PR #403, opened 2026-08-31
chore: assert the known Expo UIScene failure instead of leaving the nightly red

mahmoud-elmorabea:
Fixes MBL-2324

Independent of #400 / #401 / #402 — disjoint files, not part of that stack.

## Problem

`ios-toolchain-compatibility.yml` runs nightly and has **never** been green: 30 failure / 26 cancelled / 4 success, and those 4 predate the launch step. Both matrix legs fail at `Install and launch generated app`.

Confirmed from the launch-log artifact of [run 33368923061](https://github.com/customerio/customerio-expo-plugin/actions/runs/33368923061) (nightly, 2026-08-31):

```
failure in void _UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption(void)_block_invoke
(UIApplication_RuntimeIssues.m:106) : Application failed to launch: UIScene life cycle is required
for apps built with this SDK.
```

This is a real and known defect, and **not ours**: the Expo SDK 57 template ships no `UIApplicationSceneManifest` and no `SceneDelegate`. The plugin is already SDK-58-ready and gates scene work behind `isExpoVersion58OrHigher` (`plugin/src/ios/utils.ts`). Expo 58 exists only as a canary.

`.github/IOS-TOOLCHAIN-COMPATIBILITY.md` explicitly said these apps are expected to fail and "do not interpret it as a healthy baseline or suppress it." The trouble is that a permanently-red job is not a signal — it trains everyone to ignore the result, so a genuine regression there would go unnoticed. The answer is to **invert** the assertion, not to mute it.

## Fix

1. The launch step gains `id: launch` and `continue-on-error: true`.
2. A new step asserts **all three** of:
   - `steps.launch.outcome == 'failure'`
   - the launch log matches `UIScene life cycle is required|_UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`
   - the resolved Expo SDK major in the generated app is still `57`

   The job is GREEN while the known defect holds and RED the moment it moves, naming which of three ways:

   | flip reason | meaning |
   | --- | --- |
   | `launch-now-succeeds` | Expo's UIScene host lifecycle landed — go enable scene support and retire this gate |
   | `signature-changed` | still fails, but not with the tracked signature — a genuinely new problem |
   | `expo-sdk-moved` | no longer Expo SDK 57, so this gate's premise is stale |

3. **Slack on flip**: `if: failure() && github.event_name == 'schedule'`, using `slackapi/slack-github-action` at the *same pinned SHA* `deploy-sdk.yml` already uses (`70cd7be…`, v1.26.0), with `secrets.SLACK_WEBHOOK_URL` and `SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK`. The `schedule` guard keeps PR runs of this workflow quiet. The message names the flip reason and explains all of them.
4. `.github/IOS-TOOLCHAIN-COMPATIBILITY.md` is updated in the same PR, including the explicit line: **"Green here means the known defect is unchanged. It does NOT mean Expo works on iOS 27."** Without that the doc would read as contradicted by the new green.

## Deliberate choices

- **Gated on `steps.launch.outcome`, never on the action's `failure-reason` output.** A failing composite step does not reliably propagate its outputs, and the existing test suite already encodes that lesson with `expect(workflow).not.toContain('steps.launch.outputs.failure-reason')` — still passing.
- **`permissions: contents: read` is left unchanged.** An incoming-webhook post is a plain outbound HTTPS request with a secret; it needs no `GITHUB_TOKEN` scope. The test asserts the block was not quietly widened.
- The assertion runs **before** the artifact upload, so a flip cannot be hidden behind an upload failure. The upload keeps `if: always()`, so the log is published on every run either way.

## Verification

The assertion script was extracted from the workflow and executed against the **real** launch log from nightly run 33368923061:

| case | result |
| --- | --- |
| known failure holds (real nightly log, Expo 57.0.18) | `exit=0`, no flip reason — **the nightly would now be green** |
| launch outcome `success` | `exit=1`, `launch-now-succeeds` |
| log without the signature | `exit=1`, `signature-changed` |
| generated app on Expo 58.0.0 | `exit=1`, `expo-sdk-moved` |
| launch log missing or empty | `exit=1`, `signature-changed` |

`__tests__/workflows/ios-toolchain-compatibility.test.js` gains two cases covering the assertion shape and the notification (including that the Slack pin is read from `deploy-sdk.yml` rather than duplicated). `npx jest __tests__/workflows` → 4 passed. `actionlint` → clean on this workflow.

## Note on the brief

The failure is classified by the shared action as `did-not-survive`, not `launch-failed`: the process launches, then is terminated ~1s into the 10s survival window. The `UIScene life cycle is required` text appears in the **bounded simulator log artifact**, not in the job console — which is part of why this has been easy to skim past. Both regex alternatives match that log line. The generated app also comes from `expo-template-default@57.0.20`, not `expo-template-bare-minimum`.

### DIFF
--- a/.github/IOS-TOOLCHAIN-COMPATIBILITY.md
+++ b/.github/IOS-TOOLCHAIN-COMPATIBILITY.md
@@ -4,7 +4,7 @@
 
 The existing `Validate Plugin Compatibility` pull-request workflow remains the required stable-Xcode regression coverage for APN and FCM, and the release workflow continues to run the stable full Expo SDK matrix. Repeating those stable builds in the preview workflow would consume scarce hosted macOS capacity without adding a distinct release gate.
 
-Nightly preview failures are recorded as failed scheduled or manually dispatched runs, and this
+Nightly preview flips are recorded as failed scheduled or manually dispatched runs, and this
 repository's Actions history is the source of truth. A pull request specifically changing Expo's
 iOS toolchain integration can be
 validated before merge with a manual dispatch or an explicitly test-only workflow change. The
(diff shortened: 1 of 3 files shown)


Shahroz16 (changes requested):
Could we update the utility test with this workflow change? `Test utility functions` is currently red because the test still expects `failure() && github.event_name == 'schedule'`, while the workflow now uses `always()` plus the known-failure outcome. The Xcode 27 legs pass, but the PR is not green as-is.

I would also add `!cancelled()` to the Slack condition. With `cancel-in-progress: true`, a canceled scheduled run can leave the assertion step skipped, and `always()` can then send a false failure alert.

mahmoud-elmorabea:
Both fixed — and the `!cancelled()` point was a real bug, not just hardening.

**Test.** My fault: I changed the workflow condition and never ran the repo's suite, so the assertion at `__tests__/workflows/ios-toolchain-compatibility.test.js:101` was left asserting the old string. Updated, and strengthened to assert the intent rather than an exact match — it now also checks `failure()` is gone and `expo-version-unresolved` appears in the payload legend. All 4 tests pass locally.

**`!cancelled()`.** You're right. `cancel-in-progress: true` cancels an in-flight nightly, which leaves the assertion step `skipped` — indistinguishable under `always()` from a genuine break before the assertion, so my `skipped` branch would have alerted on every cancellation. Condition is now:

```
!cancelled() && github.event_name == 'schedule'
  && (steps.known-failure.outcome == 'failure' || steps.known-failure.outcome == 'skipped')
```

Unrelated, in case it shows in your view: `Test unit tests for plugin` was red on this branch for a while. It fails inside `install-plugin-tarball.sh`, which this PR doesn't touch — a floating resolve on a branch with no committed lockfile. A re-run with no code change is green.

mrehan27 (approved):
Inverting the permanently-red job into an asserted invariant is the right call, and naming the flip reason is what makes a real regression legible.

Checked the notification gate in all four states: a real flip still alerts if the run is cancelled afterwards, a cancelled run stays silent, and a passing gate followed by a failed upload stays silent — applying `!cancelled()` to only the `skipped` branch is what gets that right. `APP_PATH` resolves via `$GITHUB_ENV`, and the Slack pin tracks `deploy-sdk.yml` rather than duplicating it.

Shahroz16 (approved):
Looks good now. The updated tests pass, and I checked the assertion and notification cases locally: cancellation before the assertion stays quiet, a confirmed flip still alerts if cancellation happens afterwards, and an upload failure after a passing assertion stays quiet. This addresses my earlier review.

PR #406, opened 2026-09-11
feat(in-app): document the color scheme override and inbox accessibility labels

mahmoud-elmorabea:
Documents the two JavaScript-config features the React Native wrapper has added, and pins the plugin to the release that carries both. No plugin code changes — both ride the existing JavaScript config path.

- **In-app color scheme** — `inApp.colorScheme` plus the runtime `CustomerIO.inAppMessaging.setColorScheme()`, for apps whose own appearance setting can disagree with the operating system.
- **Visual inbox accessibility labels** — `inApp.notificationInboxAccessibilityLabels`, the one place the inbox still needs app-supplied strings.

Fixes MBL-2366
Fixes MBL-2479

## Why no plugin options

With native auto-initialization the SDK starts before JavaScript loads, so a later `CustomerIO.initialize()` is a no-op. Exposing either feature in `app.json` would bake the values in at prebuild: the labels would ship in a single locale, and a color scheme fixed at build time cannot follow the app's own appearance toggle. So the limitation is documented rather than half-supported.

The caveat is not identical for the two, and the README says so:

- **Labels** — unreachable under auto-initialization. An unset label leaves that element unlabeled rather than falling back to English, so an auto-initializing app simply gets no inbox labels.
- **Color scheme** — only the config option is lost. `setColorScheme()` reaches the already-initialized SDK and applies immediately, re-theming messages that are already on screen, so an auto-initializing app can still pin a variant by calling it once after startup.

## Dependency

React Native SDK **6.12.0** carries both APIs — accessibility labels from 6.11.0 (customerio/customerio-reactnative#658) and the color scheme from 6.12.0 (customerio/customerio-reactnative#660), both released. The root `peerDependencies` pin and the test app both point at that published version; no unreleased commits remain.

## Verification

- `npm ci --dry-run` clean in both the root package and `test-app`; both lockfiles resolve `customerio-reactnative` 6.12.0 from the npm registry with matching integrity hashes.
- The color scheme override was exercised on an Android emulator and an iOS simulator through the React Native sample app: the configured scheme reaches the native renderer at initialization, the runtime setter changes it live, and an invalid value is reported without disturbing the scheme already in effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,29 @@ After you add the plugin to your project, you'll need to install our React Nativ
 
 You'll find our [complete SDK documentation at https://customer.io/docs/sdk/expo](https://customer.io/docs/sdk/expo/).
 
+## Visual notification inbox accessibility labels
+
+The SDK ships no text of its own in the visual notification inbox — the empty state is an icon and the loading state is a spinner — so accessibility labels are the one place a string is still needed. Apps supply their own through `inApp.notificationInboxAccessibilityLabels` when they call `CustomerIO.initialize()` from JavaScript:
+
+```ts
+CustomerIO.initialize({
+  cdpApiKey: '...',
+  inApp: {
+    siteId: '...',
+    notificationInboxAccessibilityLabels: {
+      bell: t('inbox.bell'),
+      bellWithUnreadCount: t('inbox.unread'), // e.g. "{count} unread notifications"
+      loadingIndicator: t('inbox.loading'),
+      emptyState: t('inbox.empty'),
+    },
+  },
+});
+```
+
+**These labels require JavaScript initialization.** With native auto-initialization (a `config` block in the plugin options), the SDK is initialized before JavaScript loads, so a later `CustomerIO.initialize()` call is a no-op and the labels never reach the SDK. They are deliberately not exposed as plugin options: values in `app.json` are baked in at prebuild, which would ship a capability that only works for one locale.
+
+An unset label leaves that element unlabeled rather than falling back to English, so an auto-initializing app simply gets no inbox labels.
+
 ## Scene deep links with native auto-initialization
 
 When using Expo's scene lifecycle with Customer.io native auto-initialization, register your React Native `Linking` URL listener and then call `CustomerIO.setDeepLinkRoutingReady()`. This lets the plugin deliver URLs buffered during cold launch without requiring a second SDK initialization from JavaScript.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "customerio-expo-plugin",
-  "version": "3.7.1",
+  "version": "3.9.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "customerio-expo-plugin",
-      "version": "3.7.1",
+      "version": "3.9.1",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
@@ -36,7 +36,7 @@
         "yaml": "^2.8.3"
       },
       "peerDependencies": {
-        "customerio-reactnative": "6.9.0"
+        "customerio-reactnative": "6.10.0"
       }
     },
     "node_modules/@babel/cli": {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -18,6 +18,68 @@ After you add the plugin to your project, you'll need to install our React Nativ
...
+      bell: t('inbox.bell'),
+      bellWithUnreadCount: t('inbox.unread'), // e.g. "{count} unread notifications"
+      loadingIndicator: t('inbox.loading'),
+      emptyState: t('inbox.empty'),
+    },
+  },
+});
+```
+
+`bellWithUnreadCount` is a template: `{count}` is replaced with the number of unread messages when the bell is announced. Spell it any other way and nothing is substituted, so the screen reader reads the text verbatim, braces included, and never says the count — the SDK warns in the JavaScript console when it spots that.
```
Shahroz16: Could we say this warning is limited to development builds? `warnIf` is gated by `__DEV__`, so a release build with a malformed template will not write anything to the console.

README.md
```diff
@@ -18,6 +18,68 @@ After you add the plugin to your project, you'll need to install our React Nativ
...
+
+**These labels require JavaScript initialization.** With native auto-initialization (a `config` block in the plugin options), the SDK is initialized before JavaScript loads, so a later `CustomerIO.initialize()` call is a no-op and the labels never reach the SDK. They are deliberately not exposed as plugin options: values in `app.json` are baked in at prebuild, which would ship a capability that only works for one locale.
+
+An unset label leaves that element unlabeled rather than falling back to English, so an auto-initializing app simply gets no inbox labels.
```
Shahroz16: Could we describe the per-label fallbacks here? `bellWithUnreadCount` falls back to `bell`, and an unset loading label still leaves Android’s progress role for TalkBack while VoiceOver skips it. The current wording makes partial configuration sound like every element becomes unlabeled.

README.md
```diff
@@ -18,6 +18,68 @@ After you add the plugin to your project, you'll need to install our React Nativ
...
+```ts
+CustomerIO.inAppMessaging.setColorScheme(CioColorScheme.Light);
+```
+
+That takes effect immediately — messages already on screen, inline views included, are re-themed in
+place, so it can be called straight from the app's own appearance toggle.
+
+**Unlike the inbox accessibility labels above, native auto-initialization only affects the config
+option, not the setter.** With a `config` block in the plugin options the SDK starts before
+JavaScript loads, so `inApp.colorScheme` never arrives — but `setColorScheme()` reaches the
```
Shahroz16: Could we qualify this with `config.siteId`? The plugin only initializes MessagingInApp when `siteId` is present. Without it, both native setters log that in-app messaging is unavailable and leave the scheme unchanged, so this does not work for every auto-initialized configuration.

Shahroz16 (approved):
The dependency bump and release path look good. The remaining comments are documentation-only and can be addressed before merge.

README.md
mahmoud-elmorabea: Done in 87979e7. Confirmed — `warnIf` is `if (__DEV__ && condition)` in `src/utils/param-validation.ts`, so the warning is compiled out of release builds. The sentence now says it is a development-build diagnostic and that a malformed template ships silently otherwise.

README.md
mahmoud-elmorabea: Done in 87979e7 — you were right that the old wording made partial configuration sound worse than it is. Replaced it with a per-label table taken from the two SDKs' own declarations:

- `bell` unset → announced as an unnamed button.
- `bellWithUnreadCount` unset → falls back to `bell`, so only the count goes unannounced (the badge is always hidden from screen readers, so the count is only ever spoken through this label).
- `loadingIndicator` unset → the spinner stops being an accessibility element; VoiceOver skips it, TalkBack still reports Android's underlying progress role.
- `emptyState` unset → the icon is hidden from assistive technologies.

Kept the "no English fallback" point, which is still accurate, and dropped the implication that every element becomes unlabeled.

README.md
mahmoud-elmorabea: Done in 87979e7 — good catch, and this was the one that actually overstated behaviour.

Confirmed in both initializer templates: iOS `CustomerIOSDKInitializer.swift` guards with `if let siteId = siteId {` and Android `CustomerIOSDKInitializer.kt` with `if (!(siteId.isNullOrBlank())) {`, so the in-app messaging module is only added when a siteId is present. Without one there is no module for either path to reach — the config option is dropped as before, and the setter logs that in-app messaging is unavailable and leaves the scheme unchanged.

The claim is now qualified with `config.siteId` in the bold sentence, plus a following paragraph explaining why.

